PR #5, opened 2016-07-01
Support printing to stdout

### LINKED ISSUE #2: Support output to stdout to pipe to kubectl

janetkuo:
Fixes #2

Add a `-out` / `-stdout` flag for printing to stdout. Doesn't support printing helm chart since it doesn't make sense (or does it?). I think the naming of the flag could be better but I couldn't think of a better one for now. 

Some unsolved issues found while developing:

1. Services should be created first if possible
2. It seems output to helm chart is only supported when there's `xxx-rc.json` (not yaml, not other controllers)
3. Right now we don't stop users from creating more than one type of controllers (rc, rs, deployment, ...), but this won't work while printing to stdout and pipe to kubectl, so I changed the logic to just use the first flag specified. Does that sound good? We'll need some verification for incorrect flag usage also. 

@runseb @ngtuna

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -19,6 +19,7 @@ package app
 import (
 	"fmt"
 	"math/rand"
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


ngtuna:
Thanks @janetkuo . LGTM.

1. Yes agree, services should be created first, specially when writing them to stdout help the pipe to kubectl working better.
3. Yes, should have a verification on input flags and also printing a warning. The logic IMO would be only accepting the first flag in case of printing to stdout, otherwise keep the current support.

ngtuna:
@janetkuo I kept output to stdout option flag `--stdout` (removed `--out`) and added output to single file with flag `-o/--out` at https://github.com/skippbox/kompose2/commit/bea0e4b7c3cbccca1bf1a976416e2dec0fb1c868. They can't be declared together at the same time.

PR #28, opened 2016-07-08
Validate flags when generating charts, and prints message for file created

janetkuo:


cli/app/app.go
```diff
@@ -593,17 +597,22 @@ func ProjectKuberConvert(p *project.Project, c *cli.Context) {
 		logrus.Fatalf("Failed to parse the compose project from %s: %v", composeFile, err)
 	}
 
-	f := createOutFile(outFile)
-	defer f.Close()
+	var f *os.File
+	if !createChart {
+		f = createOutFile(outFile)
```
janetkuo: Do this because we're copying rc files to chart templates now. Can support single file later.

PR #35, opened 2016-07-13
Generate only controllers set by flag

### LINKED ISSUE #33: Default controller object is always generated.
Right now default controller is always generated. 
When I do `kompose convert --replicaset` I get Services, Deployments and ReplicaSets.
I think that this is little bit confusing (even if they are in different files).
When I run above command I would expect only Services and ReplicaSets. In most cases you need only one controller object.

When I do `kompose convert --replicaset --deployment` than I would expect both (ReplicaSets and Deployments)

kadel:
fixes #33 

After this function `ProjectKuberUp` needs to be updated, because right now it deploys only `svc` and `rc`

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -566,6 +566,11 @@ func ProjectKuberConvert(p *project.Project, c *cli.Context) {
 	createRC := c.BoolT("replicationcontroller")
 	singleOutput := len(outFile) != 0 || toStdout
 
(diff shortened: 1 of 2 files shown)


ngtuna:
LGTM. It's no problem at this moment, `ProjectKuberUp` is low priority. Enhancing `convert` function is the first serve.

PR #62, opened 2016-07-22
Add --replicas flag and changed --rc from string to bool

### LINKED ISSUE #52: Consider changing `--rc` flag to bool and adding `--replicas`
Right now all flags for specifying controller types are `bool`, except for `--rc`, which is `int`. 

This is confusing because I cannot do this 
```console
$ ../kompose convert --rc -y
Incorrect Usage.
```

And this (surprisingly) gives me a deployment 
```console
$ ../kompose convert --rc="0" -y
file "redis-svc.yaml" created
file "web-svc.yaml" created
file "redis-deployment.yaml" created
file "web-deployment.yaml" created
```

I propose that we change `--rc` back to `bool` and add another `int` flag for specifying replicas, since other controllers also take `replicas` in their spec. We can default it to `-1` as unset, and then we don't specify it in spec (and let K8s API server decide the default value). If it's set to something >= 0, we set it explicitly.

janetkuo:
Now we can do `kompose convert -rc` or `kompose convert --replicas=3`.

Fixes #52

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -276,8 +276,8 @@ func initRC(name string, service ServiceConfig, replicas int) *api.ReplicationCo
 			//Labels: map[string]string{"service": name},
 		},
 		Spec: api.ReplicationControllerSpec{
-			Replicas: int32(replicas),
 			Selector: map[string]string{"service": name},
+			Replicas: int32(replicas),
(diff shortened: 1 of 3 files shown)


janetkuo:
Got lgtm, so merging

Issue #75, opened 2016-08-02
Problems of converting volumes


janetkuo:
A few problems I discovered specific to `volumes`:

1. In kompose conversion logic, a volume is created as read-only by default. However, in Kubernetes spec, container volumeMounts is read-write by default; same as docker-compose (I believe so from looking at the file format, please correct me if not). 
2. We create `emptyDir` for volumes, but if host is specified (i.e. `HOST:CONTAINER`, see [spec](https://docs.docker.com/compose/compose-file/#/volumes-volume-driver)), shouldn't we use `hostPath` instead?
3. We create a random string (length of 20) for volume names, which it's not readable; also, volume names defined in docker-compose.yml file is ignored.

ngtuna:
Thanks @janetkuo for pointing this out. This is an important part of kompose conversion. Firstly I would say the current volume conversion was outdated a little bit. I just have a quick look at this.

1. Originally compose defined a volume is in ro mode by default (correct me if I am wrong). Now I see it set to rw mode. 
2. I believe there was something missing on volume part when I did refactoring app.go. Take a look into the original code: https://github.com/skippbox/kompose/blob/9bdc4cf2dd516590135cd12bd6bf27816f0bbffd/cli/app/app.go#L415. I will check your PR immediately.
3. Volume name is not defined in [compose v1 format](https://docs.docker.com/compose/compose-file/#/version-1). That's why I created random name. Now with v2 we have named volume. Could you add a switch/check for volume conversion if the compose file is version 1 or 2 as there are some options in volume definition of compose v2 I haven't covered yet.

janetkuo:
I looked closer at the spec and libcompose, the v2 volume you mentioned are in `Project.VolumeConfigs`, which we don't support. The logic I added in #78 is to parse `Project.ServiceConfigs.Volumes` and use the volume name specified there. 

We should add warnings for not supporting VolumeConfigs too (we don't yet).

ngtuna:
@janetkuo Can we close this issue by your merged PRs?

janetkuo:
Yes

PR #78, opened 2016-08-03
Convert volumes in [name:][host:]container[:access_mode] format

janetkuo:
#75

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -509,34 +509,68 @@ func configEnvs(name string, service ServiceConfig) []api.EnvVar {
 func configVolumes(service ServiceConfig) ([]api.VolumeMount, []api.Volume) {
 	volumesMount := []api.VolumeMount{}
 	volumes := []api.Volume{}
+	volumeSource := api.VolumeSource{}
 	for _, volume := range service.Volumes {
-		character := ":"
-		if strings.Contains(volume, character) {
-			containerDir := volume[strings.Index(volume, character)+1:]
-			containerDir = strings.TrimSpace(containerDir)
-
-			// check if ro/rw mode is defined, default rw
-			readonly := false
-			if strings.Index(volume, character) != strings.LastIndex(volume, character) {
-				mode := volume[strings.LastIndex(volume, character)+1:]
-				if strings.Compare(mode, "ro") == 0 {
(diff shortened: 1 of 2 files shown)


ngtuna:
@janetkuo did you test with compose v1 which doesn't include volume name ?

janetkuo:
Yes, I tested against `docker-gitlab.yml`, and an example output is:
```console
$ cat redisio-deployment.yaml 
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    service: redisio
  name: redisio
spec:
  replicas: 1
  selector:
    matchLabels:
      service: redisio
  strategy: {}
  template:
    metadata:
      creationTimestamp: null
      labels:
        service: redisio
    spec:
      containers:
      - image: sameersbn/redis
        name: redisio
        ports:
        - containerPort: 6379
          protocol: TCP
        resources: {}
        volumeMounts:
        - mountPath: /var/lib/redis
          name: ls6k1hh2gdnyxxvi7hvs
      restartPolicy: Always
      volumes:
      - hostPath:
          path: /srv/docker/gitlab/redis
        name: ls6k1hh2gdnyxxvi7hvs
status: {}
```

cli/app/app.go
```diff
@@ -509,36 +509,70 @@ func configEnvs(name string, service ServiceConfig) []api.EnvVar {
...
-			volumeName := RandStringBytes(20)
+		name, host, container, mode, err := parseVolume(volume)
+		if err != nil {
+			logrus.Warningf("Failed to configure container volume: %v", err)
+			continue
+		}
 
-			volumesMount = append(volumesMount, api.VolumeMount{Name: volumeName, ReadOnly: readonly, MountPath: containerDir})
+		// volumeName = random string of 20 chars
+		if len(name) == 0 {
```
janetkuo: I didn't assume that volume name is specified (even for v2, volume name isn't required), and if it's not, just use the current random volume name.

janetkuo:
Just rebased.

janetkuo:
I added a unit test, tested locally with `go test -tags experimental` (how do we enable testing in travis?)

cli/app/app.go
ngtuna: :+1:

PR #84, opened 2016-08-04
Converting compose labels to k8s annotations

### LINKED ISSUE #81: docker-compose labels should be converted to k8s annotations instead of labels
Compose labels are just metadata, unlike k8s labels, which is used to organize / categorize objects.

janetkuo:
Fixes #81

### DIFF
--- a/examples/docker-voting.yml
+++ b/examples/docker-voting.yml
@@ -4,6 +4,8 @@ services:
   vote:
     build: ./vote
     command: python app.py
+    labels:
+     - "com.example.description=Vote"
     volumes:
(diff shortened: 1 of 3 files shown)


examples/docker-voting.yml
```diff
@@ -18,6 +20,8 @@ services:
 
   db:
     image: postgres:9.4
+    labels:
+     - "com.example.description=Postres Database"
```
runseb: tiny typo there: "Postgres"

janetkuo: fixed

PR #86, opened 2016-08-04
New behavior of kompose up

ngtuna:
Fix https://github.com/skippbox/kompose/issues/40

/cc @janetkuo


cli/app/app.go
```diff
@@ -1202,80 +1209,60 @@ func print(name, trailing string, data []byte, toStdout, generateYaml bool, f *o
...
-	//		}
-	//		// submit rc to k8s
-	//		rcCreated, err := client.ReplicationControllers(api.NamespaceDefault).Create(rc)
-	//		if err != nil {
-	//			fmt.Println(err)
-	//		}
-	//		logrus.Debugf("%s\n", rcCreated)
-	//	}
-	//}
+	logrus.Infof("We are going to create deployment controller and service for your dockerized application. \n" +
```
janetkuo: Change all "Deployment controller" to "Deployment" or "Kubernetes Deployment" (when we say deployment controller we usually mean the thing that's responsible for creating/syncing Deployments).

cli/command/command.go
```diff
@@ -89,10 +89,23 @@ func ConvertCommand() cli.Command {
 func UpCommand() cli.Command {
 	return cli.Command{
 		Name:  "up",
-		Usage: "Submit rc, svc objects to kubernetes API endpoint",
+		Usage: "Deploy your Dockerized application to Kubernetes (by default create dc and svc)",
```
janetkuo: `dc` isn't an official abbreviation for deployment. Use Deployments and Services (since most kompose users wouldn't be familiar with k8s abbreviations and it's hard to Google for "what svc means").

cli/command/command.go
```diff
@@ -89,10 +89,23 @@ func ConvertCommand() cli.Command {
...
 		Action: func(c *cli.Context) {
 			app.Up(c)
 		},
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:   "file,f",
+				Usage:  fmt.Sprintf("Specify an alternate compose file (default: %s)", app.DefaultComposeFile),
```
janetkuo: "alternative"

cli/app/app.go
```diff
@@ -1260,6 +1264,8 @@ func Up(c *cli.Context) {
 		dcCreated, err := client.Deployments(api.NamespaceDefault).Create(dc)
 		if err != nil {
 			logrus.Fatalf("Failed to create deployment controller %s: ", k, err)
+		} else {
+			fmt.Println("Deployment controller " + k + " has been created.")
```
janetkuo: Consider using "%q"
```go
fmt.Printf("Deployment controller %q has been created.\n", k)
```
gives you something like `Deployment controller "web" has been created.`

cli/app/app.go
```diff
@@ -1245,6 +1247,8 @@ func Up(c *cli.Context) {
 		scCreated, err := client.Services(api.NamespaceDefault).Create(sc)
 		if err != nil {
 			logrus.Fatalf("Failed to create service %s: ", k, err)
+		} else {
+			fmt.Println("Service "+ k +" has been created.")
```
janetkuo: same here

cli/app/app.go
```diff
@@ -1202,80 +1209,60 @@ func print(name, trailing string, data []byte, toStdout, generateYaml bool, f *o
...
-	//		// submit rc to k8s
-	//		rcCreated, err := client.ReplicationControllers(api.NamespaceDefault).Create(rc)
-	//		if err != nil {
-	//			fmt.Println(err)
-	//		}
-	//		logrus.Debugf("%s\n", rcCreated)
-	//	}
-	//}
+	logrus.Infof("We are going to create deployment controller and service for your dockerized application. \n" +
+		"If you need more kind of controllers, consider to use kompose convert and kubectl")
```
janetkuo: "consider to use kompose convert and kubectl" --> "use 'kompose convert' and 'kubectl create -f' instead"

janetkuo: Should we mention that we need a running Kubernetes cluster?

cli/command/command.go
```diff
@@ -89,10 +89,23 @@ func ConvertCommand() cli.Command {
...
 	return cli.Command{
 		Name:  "up",
-		Usage: "Submit rc, svc objects to kubernetes API endpoint",
+		Usage: "Deploy your Dockerized application to Kubernetes (by default create dc and svc)",
 		Action: func(c *cli.Context) {
 			app.Up(c)
 		},
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:   "file,f",
```
janetkuo: We can set both `-f` and `--bundle` flags in a function and call that function in kompose up and kompose convert

cli/app/app.go
janetkuo: I'm thinking that we can tell the users the next step, for example `kubectl get deployment, svc` to see what's created (or `kompose ps` after we implemented it)?

janetkuo: Hm and perhaps how to delete them / cleanup?

janetkuo:
How do we handle the case that the converted k8s object is invalid (missing required fields)? Users would enter a weird state that their dockerized / kubernized apps are created partially. 

```console
$ kompose up -f docker-voting.yml 
We are going to create deployment controller and service for your dockerized application. 
If you need more kind of controllers, consider to use kompose convert and kubectl. 

WARN[0000] [worker] Service cannot be created because of missing port. 
Service result has been created.
Service vote has been created.
FATA[0000] Failed to unmarshal worker to service object: unexpected end of JSON input
```

Something to think about. We may fix it in a follow-up PR.

cli/app/app.go
janetkuo: We need to update the comment of `Up` (L1210)

janetkuo: We need the same validation in convert for checking -f and --bundle aren't specified at the same time. Make that a function and call it from both places.

cli/app/app.go
```diff
@@ -1202,80 +1209,60 @@ func print(name, trailing string, data []byte, toStdout, generateYaml bool, f *o
...
+		komposeObject = loadComposeFile(inputFile, opt)
+	}
+
+	// Convert komposeObject to K8S controllers
+	mServices, mDeployments, _, _, _, _ := komposeConvert(komposeObject, opt)
+
+	factory := cmdutil.NewFactory(nil)
+	clientConfig, err := factory.ClientConfig()
+	if err != nil {
+		logrus.Fatalf("Failed to get Kubernetes client config: %v", err)
```
janetkuo: "Failed to get Kubernetes client config" --> "Failed to get config for accessing the Kubernetes server"

ngtuna:
Thanks @janetkuo . I added a commit https://github.com/skippbox/kompose/pull/86/commits/82495e8c2b09d7602472a88a0473b20d05e5c79f based on your comments. The last comment of handling invalid converted k8s object will be fixed in a follow-up PR.

cli/app/app.go
```diff
@@ -1265,10 +1284,12 @@ func Up(c *cli.Context) {
 		if err != nil {
 			logrus.Fatalf("Failed to create deployment controller %s: ", k, err)
 		} else {
-			fmt.Println("Deployment controller " + k + " has been created.")
+			fmt.Printf("Deployment controller %q has been created.\n", k)
```
janetkuo: Deployment controller --> Deployment

cli/app/app.go
```diff
@@ -1265,10 +1284,12 @@ func Up(c *cli.Context) {
...
 		}
 		logrus.Debugf("%s\n", dcCreated)
 	}
+
+	fmt.Println("\nApplication has been deployed to Kubernetes. You can run 'kubectl get deployment, svc' for details.")
```
janetkuo: `kubectl get deployment,svc` without any space

cli/app/app.go
```diff
@@ -1207,10 +1214,27 @@ func print(name, trailing string, data []byte, toStdout, generateYaml bool, f *o
...
+
+	factory := cmdutil.NewFactory(nil)
+	clientConfig, err := factory.ClientConfig()
+	if err != nil {
+		logrus.Fatalf("Failed to get config for accessing the Kubernetes cluster: %v", err)
+	}
+	client := client.NewOrDie(clientConfig)
+
+	//FIXME: try to list out system services in order to check whether k8s is running or not
+	if err := validateK8S(client); err != nil {
```
janetkuo: You don't need this, if `factory.ClientConfig()` returns no error, k8s is running.

cli/app/app.go
```diff
@@ -1207,10 +1214,27 @@ func print(name, trailing string, data []byte, toStdout, generateYaml bool, f *o
...
 func Up(c *cli.Context) {
-	fmt.Println("We are going to create deployment controller and service for your dockerized application. \n" +
-		"If you need more kind of controllers, consider to use kompose convert and kubectl. \n")
+	fmt.Println("We are going to create Kubernetes deployment and service for your dockerized application. \n" +
+		"If you need more kind of controllers, use 'kompose convert' and 'kubectl create -f' instead. \n")
+
+	factory := cmdutil.NewFactory(nil)
+	clientConfig, err := factory.ClientConfig()
+	if err != nil {
+		logrus.Fatalf("Failed to get config for accessing the Kubernetes cluster: %v", err)
```
janetkuo: Maybe just `logrus.Fatalf("Failed to access the Kubernetes cluster: %v", err)` is enough

ngtuna:
Thanks @janetkuo. I fixed it with an 'amended' commit.

PR #91, opened 2016-08-07
loader-transformer

### LINKED ISSUE #55: Breaking code in app.go to multiple packags

ngtuna:
Fixes #55 

Added:
- loader, transformer interfaces
  + LoadFile(file string) KomposeObject
  + Transform(KomposeObject, ConvertOptions)
- bundlefile, compose packages
- kubernetes, openshift packages

@kadel Please take a review and give me your comments.


pkg/transformer/utils.go
```diff
@@ -0,0 +1,364 @@
...
+		f, err = os.Create(out)
+		if err != nil {
+			logrus.Fatalf("error opening file: %v", err)
+		}
+	}
+	return f
+}
+
+// Init RC object
+func InitRC(name string, service kobject.ServiceConfig, replicas int) *api.ReplicationController {
```
kadel: Shoulden't we move all `Init*` functions also to Kubernetes package?

kadel:
I think that it would be great to clean up ` pkg/transformer/utils.go` a little bit more (see my inline comment) But otherwise it looks good :+1: 

It would be great if @janetkuo could also check this.

kadel:
And our only unit test now fails :smiley: 
```
# github.com/skippbox/kompose/cli/app
cli/app/app_test.go:102: undefined: parseVolume
FAIL	github.com/skippbox/kompose/cli/app [build failed]
```

ngtuna:
Thanks @kadel .

Originally, I put all Init* functions into kubernetes package. However, from my viewpoint, kompose support converting input to kubernetes-based providers' primitives. So I moved all basic functions (like Init*) to transformer's utils.

ngtuna:
Well it should fail as `parseVolume()` now is moved to `transformer` package :smile_cat:

kadel:
> Originally, I put all Init* functions into kubernetes package. However, from my viewpoint, kompose support converting input to kubernetes-based providers' primitives. So I moved all basic functions (like Init*) to transformer's utils.

I don't know, for me it seems inconsistent, because you have `Init*` functions in generic `transformer` package but for example `UpdateController`, that operates on those objects, is in `kubernetes`.

For me it seems to be more logical to have everything that works with kubernetes together.

ngtuna:
> I don't know, for me it seems inconsistent, because you have Init* functions in generic transformer package but for example UpdateController, that operates on those objects, is in kubernetes.
For me it seems to be more logical to have everything that works with kubernetes together.

Okay, I agree. That would be very easy to fix. I will move them to Kubernetes except non-related k8s functions such as `print()`, `createOutFile()` or `randStringBytes()` So I will stay tuned for @janetkuo 's comments prior to get it merged.

janetkuo:
Ah, I totally missed this PR...

PR #99, opened 2016-08-11
Build statically linked binaries in makefile; remove make clean

### LINKED ISSUE #98: Release: kompose binary should be statically linked
The latest release binary is dynamically linked. We should distribute statically linked binaries instead. 
```console
$ ldd kompose 
	linux-vdso.so.1 =>  (0x00007ffe937ea000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f0a7dae5000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0a7d720000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0a7dd03000)

$ file kompose 
/usr/local/google/home/chiachenk/Downloads/kompose: ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.32, BuildID[sha1]=86c6d2ff21297a06cc7319244f35e2671612beae, not stripped
```

Setting `CGO_ENABLED=0` should do the trick.

Ref #70

janetkuo:
Fixes #98

Now `make binary`, `make all` or `make default` gives me statically linked binaries:

```console
$ ldd bundles/kompose_linux-amd64/kompose 
	not a dynamic executable
$ file bundles/kompose_linux-amd64/kompose 
bundles/kompose_linux-amd64/kompose: ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), statically linked, not stripped
```

Also removed `make clean` since the clean script doesn't exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,3 @@ all:
 
 binary:
 	CGO_ENABLED=0 ./script/make.sh binary
-
-clean:
-	./script/make.sh clean
-
--- a/Makefile
+++ b/Makefile
@@ -12,10 +12,10 @@ BIND_DIR := bundles
 default: binary
 
 all:
-	./script/make.sh
+	CGO_ENABLED=0 ./script/make.sh
 
 binary:
-	./script/make.sh binary
+	CGO_ENABLED=0 ./script/make.sh binary
 
 clean:
 	./script/make.sh clean


ngtuna:
+1 @janetkuo .

pradeepto:
@janetkuo @ngtuna  But we should have a `clean` directive, I think. Not sure how it works in Goland but, it definitely made a lot of sense in C/C++ world. We probably want something to make sure we have a pristine build. What do you all think?

janetkuo:
@pradeepto Yes we should support `make clean`, and we can add it back once we have the clean script

Issue #102, opened 2016-08-12
kompose convert panic on v1 compose file


janetkuo:
Converting examples/docker-compose.yml to k8s...
```console 
$ kompose convert 
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x8 pc=0x46be8a]
```

It's because of [this line](https://github.com/skippbox/kompose/blob/master/cli/app/app.go#L870) which references Networks (which is `nil`, since it's v1 compose file):
```go 
//FIXME: networks always contains one default element, even it isn't declared in compose v2.
if len(composeServiceConfig.Networks.Networks) > 0 &&
```

@surajssd we need to add those example files to the test

janetkuo:
Fixing in #103, still need tests

ngtuna:
Super @janetkuo . I saw this panic error last night but immediately had to go to sleep so didn't fix.Thanks for your quick react :+1:

ngtuna:
This issue also stopped me to finish my PR #91.

janetkuo:
This is fixed now

PR #103, opened 2016-08-12
Fix panic on v1 compose file

janetkuo:
#102


janetkuo:
Closing since it's fixed in HEAD

PR #104, opened 2016-08-12
Add unit test for komposeConvert

janetkuo:
Ref #34

### DIFF
--- a/pkg/kobject/kobject.go
+++ b/pkg/kobject/kobject.go
@@ -19,6 +19,7 @@ package kobject
 import (
 	"github.com/Sirupsen/logrus"
 	"github.com/fatih/structs"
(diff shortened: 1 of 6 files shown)


janetkuo:
This PR is getting bigger than I thought. PTAL and I'll add more tests in follow-up PRs.

pkg/kobject/kobject.go
```diff
@@ -162,19 +163,9 @@ type EnvVar struct {
 type Ports struct {
 	HostPort      int32
 	ContainerPort int32
-	Protocol      Protocol
+	Protocol      api.Protocol
```
ngtuna: :+1:

ngtuna:
Checked. That looks great :+1:  @janetkuo

PR #113, opened 2016-08-17
support kompose down subcommand

ngtuna:
Add new subcommand `kompose down`, in contrast with `kompose up`. There are two options:

```console
$ kompose down --file <docker-compose>
$ kompose down --bundle <bundle-file>
```
It will delete corresponding services and deployments of converted application.

```console
$ kompose down --all
Using flag --all/-a will delete all resources in the kubernetes cluster.
Are you sure to continue? (yes/no): yes
```
It will delete all resources (deployments, services, replication controllers, daemonsets) in the cluster. (only at default namespace, other namespaces are coming in follow-up PR).

Can you take a look @janetkuo ? I'm not sure if I missed something in calling kubernetes apis. For example, I set DeleteOptions = nil when deleting deployment. or ListOptions is empty. They are temporarily set until we find suitable customized options.

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -243,8 +243,7 @@ func Down(c *cli.Context) {
 		if len(file) > 0 && file != "docker-compose.yml" {
 			logrus.Fatalf("Error: --file/--bundle and --all cannot be specified at the same time")
 		}
-		fmt.Println("Using flag --all/-a will delete all resources in the kubernetes cluster.")
-		fmt.Print("Are you sure to continue? (yes/no): ")
+		fmt.Print("Are you sure? (yes/no): ")
 		if !askForConfirmation() {
 			return
 		}
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -239,18 +239,6 @@ func Down(c *cli.Context) {
 
 	validateFlags(opt, false, dabFile, inputFile)
 
-	if c.BoolT("all") {
-		if len(file) > 0 && file != "docker-compose.yml" {
-			logrus.Fatalf("Error: --file/--bundle and --all cannot be specified at the same time")
-		}
-		fmt.Print("Are you sure? (yes/no): ")
-		if !askForConfirmation() {
-			return
-		}
-		kubernetes.DeleteAll(client)
-		return
-	}
-
 	// loader parses input from file into komposeObject.
 	var l loader.Loader
 	switch inputFormat {
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -213,6 +213,62 @@ func Up(c *cli.Context) {
 	kubernetes.CreateObjects(client, objects)
 }
 
+// Down deletes all deployment, svc.
+func Down(c *cli.Context) {
+	factory := cmdutil.NewFactory(nil)
+	clientConfig, err := factory.ClientConfig()
+	if err != nil {
+		logrus.Fatalf("Failed to access the Kubernetes cluster. Make sure you have a Kubernetes running: %v", err)
+	}
+	client := client.NewOrDie(clientConfig)
+
+	inputFile := c.String("file")
+	dabFile := c.String("bundle")
+
+	komposeObject := kobject.KomposeObject{
+		ServiceConfigs: make(map[string]kobject.ServiceConfig),
+	}
+
+	file := inputFile
+	if len(dabFile) > 0 {
+		inputFormat = "bundle"
+		file = dabFile
+	}
+
+	opt := kobject.ConvertOptions{}
+
+	validateFlags(opt, false, dabFile, inputFile)
+
+	if c.BoolT("all") {
+		if len(file) > 0 && file != "docker-compose.yml" {
+			logrus.Fatalf("Error: --file/--bundle and --all cannot be specified at the same time")
+		}
+		fmt.Println("Using flag --all/-a will delete all resources in the kubernetes cluster.")
+		fmt.Print("Are you sure to continue? (yes/no): ")
+		if !askForConfirmation() {
+			return
+		}
+		kubernetes.DeleteAll(client)
+		return
+	}
+
+	// loader parses input from file into komposeObject.
+	var l loader.Loader
+	switch inputFormat {
+	case "bundle":
+		l = new(bundle.Bundle)
+	case "compose":
+		l = new(compose.Compose)
+	default:
+		logrus.Fatalf("Input file format is not supported")
+	}
+	komposeObject = l.LoadFile(file)
+
+	for k := range komposeObject.ServiceConfigs {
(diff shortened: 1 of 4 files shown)


ngtuna:
Fixes https://github.com/skippbox/kompose/issues/41

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -350,3 +350,75 @@ func CreateObjects(client *client.Client, objects []runtime.Object) {
...
 }
+
+func DeleteObjects(client *client.Client, name string) {
+	err := client.Services(api.NamespaceDefault).Delete(name)
+	if err != nil {
+		logrus.Fatalf("Error: '%v' while deleting service: %s", err, name)
+	}
+	logrus.Infof("Successfully deleted service: %s", name)
+
+	err = client.Deployments(api.NamespaceDefault).Delete(name, nil)
```
kadel: This is not enough to just delete Deployments, because it will leave running pods behind.
First you need to scale Deployments to 0, and than it is safe to delete Deployment.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -350,3 +350,75 @@ func CreateObjects(client *client.Client, objects []runtime.Object) {
 	}
 	fmt.Println("\nApplication has been deployed to Kubernetes. You can run 'kubectl get deployment,svc' for details.")
 }
+
+func DeleteObjects(client *client.Client, name string) {
+	err := client.Services(api.NamespaceDefault).Delete(name)
+	if err != nil {
+		logrus.Fatalf("Error: '%v' while deleting service: %s", err, name)
```
kadel: Rather then failing after fist error, we should just display error, and continue to try delete next object.
Could we return this error, and handle it in `func Down(c *cli.Context)`?

pkg/transformer/kubernetes/kubernetes.go
kadel: Hmm, it seems that we need also somehow clean ReplicaSets :-(

janetkuo: Yeah, in the future k8s support server-side cascading deletion. For now we can use reapers.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -350,3 +350,75 @@ func CreateObjects(client *client.Client, objects []runtime.Object) {
...
+	if err != nil {
+		logrus.Fatalf("Error: '%v' while deleting deployment: %s", err, name)
+	}
+	logrus.Infof("Successfully deleted deployment: %s", name)
+}
+
+func DeleteAll(client *client.Client) {
+	//delete all svc
+	listOpts := api.ListOptions{}
+	svcs, err := client.Services(api.NamespaceDefault).List(listOpts)
```
janetkuo: Flag help says delete everything in all namespaces but here only default namespace is referenced

cli/app/app.go
```diff
@@ -365,3 +393,19 @@ func sortServicesFirst(objs *[]runtime.Object) {
...
 	*objs = ret
 }
+
+func askForConfirmation() bool {
+	var response string
+	_, err := fmt.Scanln(&response)
+	if err != nil {
+		logrus.Fatal(err)
+	}
+	if response == "yes" {
```
janetkuo: Is this case sensitive? Is "y/n" supported also? Should we default to "no"?

cli/app/app.go
```diff
@@ -365,3 +393,19 @@ func sortServicesFirst(objs *[]runtime.Object) {
...
+		return true
+	} else if response == "no" {
+		return false
+	} else {
+		fmt.Println("Please type yes or no and then press enter:")
```
janetkuo: I'm thinking we can also print everything that's going to be deleted. Something like "Here are the resources to be deleted. Are you sure (yes/NO)?"

pkg/transformer/kubernetes/kubernetes.go
ngtuna: That's a good idea. Thank @kadel

pkg/transformer/kubernetes/kubernetes.go
ngtuna: OK thanks @janetkuo & @kadel. Let me check reapers and make a follow-up commit.

cli/app/app.go
ngtuna: That makes sense. Will add a follow-up commit.

cli/app/app.go
ngtuna: I'm not sure, but yes/no is enough IMO.

pkg/transformer/kubernetes/kubernetes.go
ngtuna: :+1: fixed in commit amend

pkg/transformer/kubernetes/kubernetes.go
ngtuna: Need confirmation @janetkuo / @kadel : We can use:
- ServiceReaper to delete svc
- DeploymentReaper to delete pod, rc, replicaset & deployment
- DaemonReaper to delete daemonset
- JobReaper to delete job. 

Am I missing or wrong something?

kadel: You are right.

btw: look at  [`func ReaperFor(....`](https://github.com/kubernetes/kubernetes/blob/v1.3.5/pkg/kubectl/stop.go#L64) this might  useful.

Let me know if I can help you with that more.

ngtuna: I added a follow-up commit below. Can you check it?

ngtuna: I'm not sure we should delete all jobs or not. Marked it //TODO

pkg/transformer/kubernetes/kubernetes.go
ngtuna: Fixed by https://github.com/skippbox/kompose/pull/113/commits/66601088e10053a58bf544ecd3f30000abfa90c7

ngtuna:
Stuck at https://github.com/skippbox/kompose/issues/117 where I can't use "k8s.io/kubernetes/pkg/kubectl/cmd" for printing resources in format of `kubectl get`

kadel:
> Stuck at #117 where I can't use "k8s.io/kubernetes/pkg/kubectl/cmd" for printing resources in format of kubectl get

https://github.com/skippbox/kompose/issues/117#issuecomment-242029705

kadel:
@ngtuna  If it is just for printing, can we use something else for now?

ngtuna:
@kadel yes I'm also thinking about it

PR #118, opened 2016-08-22
Remove hostPath and print warnings

### LINKED ISSUE #109: hostPath volumes?
We use `hostPath` in volumes?
This doesn't really makes sense to me. When you are deploying to Kubernets cluster `hostPath` will **never** contain files that you have on your local machine from which you are running `kompose`. Unless you somehow copy them manually.
Another thing with `hostPath` is that even if you somehow upload you files to it, whenever pod is rescheduled to another node `hostPath` can contain completely different files.

Because of above I think that `hostPath` can confuse users instead of helping.

Ideally user should  be able to choose which type of volume he/she wants (this will usually depend on his environment setup (gcePersistentDisk, awsElasticBlockStore, nfs ....)).

For now I propose using emptyDir, and displaying warning or some message explaining why this is. And that right now you can't copy content of local directory to cluster volume.

janetkuo:
Fixes #109

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -192,10 +192,9 @@ func ConfigVolumes(service kobject.ServiceConfig) ([]api.VolumeMount, []api.Volu
 		volumesMount = append(volumesMount, api.VolumeMount{Name: name, ReadOnly: readonly, MountPath: container})
 
 		if len(host) > 0 {
-			volumeSource = api.VolumeSource{HostPath: &api.HostPathVolumeSource{Path: host}}
-		} else {
-			volumeSource = api.VolumeSource{EmptyDir: &api.EmptyDirVolumeSource{}}
+			logrus.Warningf("Volume mount on the host %q isn't supported - ignoring path on the host", host)
 		}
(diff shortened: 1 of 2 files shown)


pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -192,10 +192,9 @@ func ConfigVolumes(service kobject.ServiceConfig) ([]api.VolumeMount, []api.Volu
 		volumesMount = append(volumesMount, api.VolumeMount{Name: name, ReadOnly: readonly, MountPath: container})
 
 		if len(host) > 0 {
-			volumeSource = api.VolumeSource{HostPath: &api.HostPathVolumeSource{Path: host}}
-		} else {
-			volumeSource = api.VolumeSource{EmptyDir: &api.EmptyDirVolumeSource{}}
+			logrus.Warningf("Volume mount on the host %q isn't supported - ignoring path on the host", host)
```
ngtuna: I'm thinking we should make a more meaningful message. Something like @kadel mentioned

kadel: This message is great. That's what I had in my mind.

ngtuna: Then I'm okay

PR #127, opened 2016-08-29
docker-compose - Entrypoint support

kadel:
We had Command and Args together with Entrypoint in Kompose object.
This doesn't make much sense, as Entrypoint and Command are same thing.
I've removed Entrypoint from Kompose object in favor of Command to keep
same naming convensions as in Kubernetes.

Entrypoint from docker-compose.yml is now saved to Command and Command
is saved to Args (http://kubernetes.io/docs/user-guide/containers/).

docker-compose.yml | Kompose object
---------------------------|----------------------
    Entrypoint              |    Command
    Command             |     Args

### DIFF
--- a/pkg/loader/compose/compose.go
+++ b/pkg/loader/compose/compose.go
@@ -151,8 +151,8 @@ func (c *Compose) LoadFile(file string) kobject.KomposeObject {
 			serviceConfig := kobject.ServiceConfig{}
 			serviceConfig.Image = composeServiceConfig.Image
 			serviceConfig.ContainerName = composeServiceConfig.ContainerName
-			serviceConfig.Entrypoint = composeServiceConfig.Entrypoint
-			serviceConfig.Command = composeServiceConfig.Command
+			serviceConfig.Command = composeServiceConfig.Entrypoint
+			serviceConfig.Args = composeServiceConfig.Command
 
 			// load environment variables
 			envs := loadEnvVars(composeServiceConfig.Environment.ToMap())
--- a/pkg/loader/compose/compose.go
+++ b/pkg/loader/compose/compose.go
@@ -119,6 +119,7 @@ func (c *Compose) LoadFile(file string) kobject.KomposeObject {
 
 	// load compose file into composeObject
 	composeObject := project.NewProject(&context.Context, nil, nil)
+
 	err := composeObject.Parse()
 	if err != nil {
 		logrus.Fatalf("Failed to load compose file: %v", err)
@@ -150,6 +151,8 @@ func (c *Compose) LoadFile(file string) kobject.KomposeObject {
 			serviceConfig := kobject.ServiceConfig{}
 			serviceConfig.Image = composeServiceConfig.Image
 			serviceConfig.ContainerName = composeServiceConfig.ContainerName
+			serviceConfig.Entrypoint = composeServiceConfig.Entrypoint
+			serviceConfig.Command = composeServiceConfig.Command
 
 			// load environment variables
 			envs := loadEnvVars(composeServiceConfig.Environment.ToMap())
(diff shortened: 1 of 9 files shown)


kadel:
Heyo, can please someone review this?
@surajssd @ngtuna @janetkuo ?

runseb:
@kadel I think it is a bit more subtle than this. 

I haven't checked but I believe Kubernetes cmd/args overwrite the Dockerfile entrypoint.

you could have a compose file with a build instructions that also specifies a legitimate default entry point which would be different than the compose CMD being passed.

I maybe wrong though, but I would like to see some examples and manual testing.

what do you think ?

kadel:
@runseb I'm afraid that I'm not sure if I understand :-( 
Kubernetes can overwrite cmd/args from image and docker-compose is doing the same. 
It is used to overwrite default cmd/args from docker image.
What is in docker-compose.yml overwrites what is specified in image.

runseb:
so the command spec in a Pod overwrites the entrypoint of an Image ?

runseb:
ok I think I get it.

in compose entrypoint and command overwrite entrypoint and command :)
but in k8s the Pod does this by overwriting the entry point using what is in command, and overwrites command with what is in args .

kadel:
I will try to show some examples how I believe it should work.

I have image foo/bar that is build from this:
```
.....
ENTRYPOINT entrImg
CMD  cmdImg
```

compose file like this 
```yml
services:
  foo:
    image: foo/bar
```
should run `foo/bar entrImg cmdImg`
in k8s PodSpec it should be:
```json
"spec": {
  "containers":  [  
    {"image": "foo/bar"}
   ]
}
```

docker-compose.yml
```yml
services:
  foo:
    image:  foo/bar
    commnad: cmdCompose

```
should run `foo/bar entrImg cmdCompose`
podSpec:
```json
"spec": {
  "containers":  [  
    {"image": "foo/bar",
      "args": ["cmdCompose"]}
   ]
}
```


docker-compose.yml
```yml
services:
  foo:
    image:  foo/bar
    entrypoint: entrCompose

```
should run `foo/bar entrCompose cmdImage`
podSpec:
```json
"spec": {
  "containers":  [  
    {"image": "foo/bar",
      "command": ["entrCompose"]}
   ]
}
```


docker-compose.yml
```yml
services:
  foo:
    image:  foo/bar
    entrypoint: entrCompose
    command: cmdCompose

```
should run `foo/bar entrCompose cmdCompose`
podSpec:
```json
"spec": {
  "containers":  [  
    {"image": "foo/bar",
      "command": ["entrCompose"],
      "args": ["cmdCompose"]}
   ]
}
```

runseb:
thanks

+1

kadel:
> in compose entrypoint and command overwrite entrypoint and command :)
> but in k8s the Pod does this by overwriting the entry point using what is in command, and overwrites command with what is in args .

exactly 

k8s names are different,  it uses same name for different things :-D 
[Here](http://kubernetes.io/docs/user-guide/containers/) is k8s documentation for what is tried to explain above :-D.

runseb:
yeah, I think it is still a bit subtle because if an image does not have an entry point defined, it will use sh -c

and you can actually specify the command simply in args parameter of a pod...

confusing IMHO but nothing to do with your PR.

janetkuo:
Code looks good. Maybe add a test in test-cmd

pkg/loader/compose/compose.go
```diff
@@ -150,6 +151,8 @@ func (c *Compose) LoadFile(file string) kobject.KomposeObject {
 			serviceConfig := kobject.ServiceConfig{}
 			serviceConfig.Image = composeServiceConfig.Image
 			serviceConfig.ContainerName = composeServiceConfig.ContainerName
+			serviceConfig.Command = composeServiceConfig.Entrypoint
+			serviceConfig.Args = composeServiceConfig.Command
```
ngtuna: :+1:

kadel:
I've added tests, Merging now.

PR #135, opened 2016-08-31
New guestbook example

runseb:
this adds the guestbook example.

It is really sweet, it works with docker-compose (v2) on a Docker for Mac, and straight up with `kompose up` on minikube.

### DIFF
--- a/examples/docker-guestbook.yml
+++ b/examples/docker-guestbook.yml
@@ -1,0 +1,19 @@
+version: "2"
+
+services:
+  redis-master:
+    image: gcr.io/google_containers/redis:e2e 
+    ports:
+      - "6379"
+  redis-slave:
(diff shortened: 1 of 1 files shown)


ngtuna:
Great. Merge it.

pydevops:
@runseb hi, just wonder how to expose the port 80 in minikube outside the vm. The kompose generates "loadBalancer "service load only while minikube can only expose port via NodePort.

runseb:
@pydevops kompose should generate a cluster IP service. To access the app running in minikube you need to use 'kubectl proxy' or edit the service manually to become a nodePort ....or run the ingress controller in minikube and create an ingress.

check this blog:
http://www.skippbox.com/using-your-docker-compose-files-with-kubernetes/

PR #137, opened 2016-09-02
update README with guestbook sample app

ngtuna:
Add guestbook sample to README. 
/cc @runseb


runseb:
I would move this to docs/user-guide.md

we should keep the main README relatively clean for the proposal review to join the incubator.

ngtuna:
Ah... OK so I should close this one and make a new PR adding guestbook to user-guide.md and also including a simple guideline for `kompose down`

PR #141, opened 2016-09-03
update user guide: add `kompose up`, `kompose down`

ngtuna:
small changes to user guide:
- kompose up, kompose down


runseb:
+1

PR #148, opened 2016-09-08
[WIP] Add Socks Shop example

sebgoa:
This is in progress.

Socks Shop is an example microservices applications that weave works uses to demo their products.
It is a fairly complete docker-compose file.

We should make sure kompose can convert it, it is a good example.

### DIFF
--- a/examples/docker-sockshop.yml
+++ b/examples/docker-sockshop.yml
@@ -1,0 +1,171 @@
+version: "2"
+
+services:
+  front-end:
+    image: weaveworksdemos/front-end
+    hostname: front-end
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - external
+      - secure
+      - internal
+  edge-router:
+    image: weaveworksdemos/edge-router
+    ports:
+      - '80:80'
+    hostname: edge-router
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - external
+  catalogue:
+    image: weaveworksdemos/catalogue
+    hostname: catalogue
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - external
+  catalogue-db:
+    image: mysql
+    hostname: catalogue-db
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+      - MYSQL_ROOT_PASSWORD=${MYSQL_ROOT_PASSWORD}
+      - MYSQL_ALLOW_EMPTY_PASSWORD=true
+      - MYSQL_DATABASE=socksdb
+    networks:
+      - external
+  cart:
+    image: weaveworksdemos/cart
+    hostname: cart
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - internal
+  cart-db:
+    image: mongo
+    hostname: cart-db
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - internal
+  orders:
+    image: weaveworksdemos/orders
+    hostname: orders
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - internal
+      - secure
+      - backoffice
+  orders-db:
+    image: mongo
+    hostname: orders-db
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - internal
+  shipping:
+    image: weaveworksdemos/shipping
+    hostname: shipping
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - backoffice
+  queue-master:
+    image: weaveworksdemos/queue-master
+    hostname: queue-master
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - backoffice
+  rabbitmq:
+    image: rabbitmq:3
+    hostname: rabbitmq
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - backoffice
+  payment:
+    image: weaveworksdemos/payment
+    hostname: payment
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - reschedule=on-node-failure
+    networks:
+      - secure
+  user:
+    image: weaveworksdemos/user
+    hostname: user
+    dns: 172.17.0.1
+    dns_search: weave.local
+    restart: always
+    environment:
+      - MONGO_HOST=user-db:27017
+      - reschedule=on-node-failure
+    networks:
+      - secure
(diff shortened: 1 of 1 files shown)


runseb:
Right now I get this error:

```
$ ./kompose convert -f ./examples/docker-sockshop.yml
WARN[0000] Unsupported network configuration of compose v2 - ignoring 
WARN[0000] Unsupported key dns - ignoring               
WARN[0000] Unsupported key dns_search - ignoring        
WARN[0000] Unsupported key hostname - ignoring          
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0x1205580, 0x8223ca050)
	/usr/local/go/src/runtime/panic.go:464 +0x3e6
github.com/skippbox/kompose/vendor/github.com/docker/libcompose/yaml.toMap(0x82286be30, 0x1, 0x1, 0x1481a70, 0x1, 0x40)
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/vendor/github.com/docker/libcompose/yaml/types_yaml.go:223 +0x187
github.com/skippbox/kompose/vendor/github.com/docker/libcompose/yaml.(*MaporEqualSlice).ToMap(0x822428150, 0x822428000)
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/vendor/github.com/docker/libcompose/yaml/types_yaml.go:126 +0x4b
github.com/skippbox/kompose/pkg/loader/compose.(*Compose).LoadFile(0x1f9fc78, 0x82262f3e0, 0x1e, 0x0)
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/pkg/loader/compose/compose.go:155 +0x91d
github.com/skippbox/kompose/cli/app.Convert(0x822567cc0)
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/cli/app/app.go:146 +0x4e8
github.com/skippbox/kompose/cli/command.ConvertCommand.func1(0x822567cc0)
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/cli/command/command.go:32 +0x21
github.com/skippbox/kompose/vendor/github.com/urfave/cli.Command.Run(0x1486fc0, 0x7, 0x0, 0x0, 0x0, 0x0, 0x0, 0x82252a9b0, 0x4b, 0x0, ...)
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/vendor/github.com/urfave/cli/command.go:179 +0x1399
github.com/skippbox/kompose/vendor/github.com/urfave/cli.(*App).Run(0x8225326e0, 0x8223c0400, 0x4, 0x4, 0x0, 0x0)
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/vendor/github.com/urfave/cli/app.go:196 +0x137c
main.main()
	/Users/sebastiengoasguen/Documents/go/src/github.com/skippbox/kompose/cli/main/main.go:47 +0x337
```

runseb:
thanks. We do need to check it this works OOTB with docker-compose.

kadel:
hmm that is strange, for me it fails with docker-compose
```
▶ docker-compose -f  docker-sockshop.yml up
WARNING: The MYSQL_ROOT_PASSWORD variable is not set. Defaulting to a blank string.
ERROR: The Compose file './docker-sockshop.yml' is invalid because:
services.user.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.catalogue-db.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.cart.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.cart-db.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.catalogue.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.edge-router.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.front-end.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.orders.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.orders-db.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.payment.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.queue-master.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.rabbitmq.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.shipping.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
services.user-db.environment contains {"reschedule": "on-node-failure"}, which is an invalid type, it should be a string
```

kadel:
Oh noooo, sorry. My bad.
It really works :-) I have modified file with space. And that fails with docker-compose.
You are right your original file works with docker-compose.

kadel:
In that case, something is wrong  in Kompose

runseb:
here is the source https://github.com/microservices-demo/microservices-demo

examples/docker-sockshop.yml
```diff
@@ -0,0 +1,171 @@
...
+  user-sim:
+    image: weaveworksdemos/load-test
+    hostname: user-simulator
+    domainname: weave.local
+    networks:
+      - external
+    command: "-d 60 -r 200 -c 2 -h edge-router"
+networks:
+  external:
+    driver: weavemesh
```
runseb: Had to remove this driver spec on my laptop (don't have weave plugin) to get it to work with docker-compose.

runseb:
ok so if we switch to:

```
    environment:
      - reschedule=on-node-failure
```

Then it works just fine (no need for space).

The faulty code seems to be in:
https://github.com/skippbox/kompose/blob/6cbdec444b06d9c0905b7c6fcc03b7b849b15a29/vendor/github.com/docker/libcompose/yaml/types_yaml.go#L124

runseb:
I think this is actually a docker-compose problem.
Docker-compose parses the yaml without errors and starts the containers but the environment variable does not get set.

for example if we look at this service:

```
  user:
    image: weaveworksdemos/user
    hostname: user
    dns: 172.17.0.1
    dns_search: weave.local
    restart: always
    environment:
      - MONGO_HOST=user-db:27017
      - reschedule:on-node-failure
```

the set environment looks like

```
$ docker exec -ti c74c4846b571 env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=user
MONGO_HOST=user-db:27017
no_proxy=*.local,169.254/16
HATEAOS=user
USER_DATABASE=mongodb
HOME=/root
```

So the `reschedule` env is not set. If we use equal sign it gets set.

kadel:
docker-compose in 1.8 version is still in Python and doesn't use libcompose.
This might be reason why it gets silently ignored there and Kompose panics.

I'm confused from all this. If this supposed to set `reschedule` env it should be like this:
```
environment:
  reschedule: on-node-failure
```
(environment as dict so no `-`)
```
environment:
  - reschedule=on-node-failure
```
(environment as array)


 I think that `reschedule:on-node-failure` is wrong and source of this is swarm [documentation](https://docs.docker.com/swarm/scheduler/rescheduling/#/rescheduling-policies).
They say: "You can do this with the `reschedule` environment variable "
and later on again: "To set the on-node-failure policy with a reschedule environment variable:
```
$ docker run -d -e reschedule:on-node-failure redis
```
"
But his command doesn't set reschedule env variable.
```
▶ docker run -it -e reschedule:on-node-failure  busybox env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=a1ce60e6b16d
TERM=xterm
HOME=/root
```

There should be `=` instead of `:`

runseb:
agreed. and I filed a bug with the microservices demo guys to let them know...I don't think they tested this.

ngtuna:
Investigating the compose file, I found that currently we can't make the sock shop app running on kubernetes via kompose. The reason is that there are some services have not been created due to lack of ports configuration in the compose file, so "micro-services" can not find each other. Running the app with `docker-compose` would be very handy but `kompose` at this moment can not detect opened ports in images metadata until we resolve https://github.com/skippbox/kompose/issues/146

cdrage:
@sebgoa Should we close this PR?

cdrage:
Hey @sebgoa I'm just going to close this PR for now :)

PR #171, opened 2016-09-26
remove executable perms from docs

dustymabe:


dustymabe:
having docs that are executable throws off some red flags in some downstream package management tools, this PR moves them to not be executable.

sebgoa:
my bad.

I will merge as soon as the travis runs finishes

dustymabe:
> my bad.

no need to apologize at all :) 

> 
> I will merge as soon as the travis runs finishes

+1

PR #178, opened 2016-09-29
Add .gitignore for Go files + compiled Kompose file

cdrage:
Adds .gitignore for common Go-related files as well as ignoring the
binary generate via go build and/or `make binary`.


sebgoa:
fair enough.

merging

PR #188, opened 2016-10-06
fix param order in readme

sharpner:
tried the latest stable (mac os x) and I had to use ``` kompose convert docker-compose.yml```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ file "web-svc.json" created
 file "web-deployment.json" created
 file "redis-deployment.json" created
 
-$ kompose docker-compose.yml convert
+$ kompose convert docker-compose.yml
 WARN[0000]: Unsupported key networks - ignoring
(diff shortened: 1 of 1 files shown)


sebgoa:
thanks for reporting the issue. There is a `-f` missing in the readme that's for sure.

But the order is correct for HEAD.

I suspect you download the 0.1.0 release ?

We will put a 0.1.1 release up to avoid this. I will keep the PR open until we fix it. Hopefully today.

@ngtuna, can you upload a 0.1.1 release in github releases page.

sharpner:
ah yeah I downloaded 0.1.0. Thanks for the fast response :)

ngtuna:
@sharpner 0.1.1 is made for you https://github.com/skippbox/kompose/releases/tag/v0.1.1

sharpner:
perfect thx

PR #211, opened 2016-10-17
remove unknown args and added tests

### LINKED ISSUE #193: error on extraneous/unexpected cli input
we should probably error out if we get extra arguments/input on the CLI that we don't expect or understand. One example of this is when I was trying to start the gitlab example but I was doing it wrong:

```
[vagrant@f24 examples (upstreammaster=)]$ ../kompose --version
kompose version 0.1.1 (5a37bf0)
[vagrant@f24 examples (upstreammaster=)]$ ../kompose up docker-gitlab.yml 

INFO[0000] Successfully created service: web            
INFO[0000] Successfully created service: redis          
INFO[0000] Successfully created deployment: web         
INFO[0000] Successfully created deployment: redis       

```

So you can see I was trying to run gitlab, but I didn't specify things like I should have: `kompose -f docker-gitlab.yml up`. This means that it actually ignored the `docker-gitlab.yml` that I passed on the command line and used the default `docker-compose.yml` instead and it looks like the application is brought up (if you aren't paying attention too close). 

Now if instead it would error out when it detected extraneous input then I would have found out what was wrong sooner than later.

procrypt:
Fix #193 
cc @kadel @dustymabe @sebgoa @surajssd

### DIFF
--- a/cli/app/app.go
+++ b/cli/app/app.go
@@ -74,6 +74,10 @@ func validateFlags(c *cli.Context, opt *kobject.ConvertOptions) {
 	if len(dabFile) > 0 && len(opt.InputFile) > 0 && opt.InputFile != DefaultComposeFile {
 		logrus.Fatalf("Error: 'compose' file and 'dab' file cannot be specified at the same time")
 	}
+
+	if len(c.Args()) != 0 {
+		logrus.Fatal("Unknown Argument(s): ", strings.Join(c.Args(), ","))
+	}
 }
 
 func validateControllers(opt *kobject.ConvertOptions) {
--- a/script/test/cmd/tests.sh
+++ b/script/test/cmd/tests.sh
@@ -58,4 +58,10 @@ convert::expect_success "kompose -f $KOMPOSE_ROOT/script/test/fixtures/volume-mo
 # openshift test
 convert::expect_success "kompose --provider=openshift -f $KOMPOSE_ROOT/script/test/fixtures/volume-mounts/simple-vol-mounts/docker-compose.yml convert --stdout --dc" "$KOMPOSE_ROOT/script/test/fixtures/volume-mounts/simple-vol-mounts/output-os.json"
 
+######
+# Tests related to unknown arguments with cli commands
+convert::expect_failure "kompose up $KOMPOSE_ROOT/script/test/fixtures/gitlab/docker-compose.yml" "Unknown Argument docker-gitlab.yml"
+convert::expect_failure "kompose down $KOMPOSE_ROOT/script/test/fixtures/gitlab/docker-compose.yml" "Unknown Argument docker-gitlab.yml"
+convert::expect_failure "kompose convert $KOMPOSE_ROOT/script/test/fixtures/gitlab/docker-compose.yml" "Unknown Argument docker-gitlab.yml"
+
 exit $EXIT_STATUS


dustymabe:
can you rebase now that the rename has happened?

procrypt:
@dustymabe rebase done.

dustymabe:
@procrypt - can you "sign the cla"?  basically create a [linuxfoundation account](https://www.linuxfoundation.org/caslogin)  with your redhat email address and then visit [this link](https://identity.linuxfoundation.org/projects/cncf) and "sign up as an employee".

cli/app/app.go
```diff
@@ -253,3 +256,10 @@ func askForConfirmation() bool {
 		return askForConfirmation()
 	}
 }
+
+func unknownArgs(c *cli.Context) {
```
dustymabe: my preference would be to rename this function to `TestForUnknownArgs` or `CheckForUnknownArgs`

kadel:
Just one thing. Have you consider putting that check to  `validateFlags` instead of adding it to every command?
`validateFlags` is already doing some validation and it is already called in every command.

I'm not saying that you should change it like this. I'm just interested why you did it like this. Valid answer can be that you think that it makes more sense the way you did it. :wink:

procrypt:
@kadel I didn't know about the function `validateFlags` we have for doing the validation thing, so I went ahead and wrote a new function which does the validation :)
I can update the PR if you want to use `validateFlags` instead of the `CheckForUnknownArgs`.

kadel:
I think that would be better to have all the validation on one place ;-)

procrypt:
@kadel I'll update the PR :)

kadel:
Oh and one more think. It looks like your commit has some weird Author.
`Author: Abhishek <abhishek@dhcp35-217.lab.eng.blr.redhat.com>` so even if you already signed CLA it will still fail because it cannot associate this commit with your github account.

procrypt:
I'll check and make sure it will be correct this time.

script/test/cmd/tests.sh
```diff
@@ -58,4 +58,10 @@ convert::expect_success "kompose -f $KOMPOSE_ROOT/script/test/fixtures/volume-mo
 # openshift test
 convert::expect_success "kompose --provider=openshift -f $KOMPOSE_ROOT/script/test/fixtures/volume-mounts/simple-vol-mounts/docker-compose.yml convert --stdout --dc" "$KOMPOSE_ROOT/script/test/fixtures/volume-mounts/simple-vol-mounts/output-os.json"
 
+######
+# Tests related to unknown arguments with cli commands
+convert::expect_failure "kompose up docker $KOMPOSE_ROOT/script/test/fixtures/unknown-arguments/docker-gitlab.yml" "Unknown Argument docker-gitlab.yml"
```
sebgoa: 'kompose up docker' I am not sure what that is ? Why do you add _docker_ in that line ?

cli/app/app.go
```diff
@@ -203,6 +205,7 @@ func Up(c *cli.Context) {
 
 // Down deletes all deployment, svc.
 func Down(c *cli.Context) {
+	CheckForUnknownArgs(c)
```
sebgoa: Did you want to update this function and actually use `ValidateFlags` as mentioned by @kadel

script/test/cmd/tests.sh
procrypt: @sebgoa This might added by mistake, will make the changes.

cli/app/app.go
procrypt: @sebgoa I'm working on this will update the PR.

procrypt:
@kadel @sebgoa I have update the PR, can you please review it.

procrypt:
@kadel :)

pradeepto:
Is this merge able now? cc @dustymabe  @procrypt @kadel @ngtuna

kadel:
@pradeepto Yes it is. 
Some time ago we established workflow where everyone is merging his/her own PR after LGTM, so I left this for @procrypt  to merge it himself.

procrypt:
@kadel I was not aware of that sorry my bad.

kadel:
Ah, no don't worry.
We should probably talk about this and document PR review process, and change it, because this cant' work for someone who doesn't have write access to repo.

kadel:
Ah, no don't worry.
We should probably  talk about this and document PR review process.

PR #217, opened 2016-10-18
example: add mlbparks example

dustymabe:


kadel:
maybe we should add volume to mongodb for this to work on OpenShift online:
```
    volumes:
      - /var/lib/mongodb/data
```

more info about why this is required https://bugzilla.redhat.com/show_bug.cgi?id=1318542

dustymabe:
I had a volume in there but I took it out because if you don't have persistent volumes set up on your test cluster then the thing will fail to come up. I can add it back.

kadel:
@dustymabe hmm, that is good point. So now we have to choose if it should by default work on OpenShift online or on single node test cluster.

Can we maybe have volumes there but commented out?

sebgoa:
will probably need to organize examples better.

But for this, if it works on minikube, I saw we merge.  I will test it right now.

sebgoa:
Ok I tested this and it works fine on minikube.

However the image mlbparks is really big ~1GB and I am not sure what it brings more than the guestbook app.

dustymabe:
hey @sebgoa - I actually didn't realize it was so big. it is based on the `openshift/wildfly-100-centos7` image that is pretty large. I like this app because it is flashy but we don't have to include it as it is a large example.

If we do decide to include it I added the volume as tkral suggested. Let me know what you want to do.

sebgoa:
@dustymabe yeah, I think the image is too big...we should only include examples that we can do during a talk. Also the map has a large "openshift" statement in it. I'd rather see _kubernetes_

dustymabe:
closing

PR #240, opened 2016-10-26
Make OpenShift inherit from Kubernetes

dustymabe:


dustymabe:
This is just a first stab and not ready yet. @kadel we really need to refactor k8sutils.go and pull out some of those functions into a separate file I think. Help me find out which functions should go in what file.

cli/app/app.go
```diff
@@ -242,6 +228,21 @@ func Down(c *cli.Context) {
 
 }
 
+// Convenience method to return the appropriate Transformer based on
+// what provider we are using.
+func getTransformer(opt kobject.ConvertOptions) transformer.Transformer {
```
kadel: can be this moved to transformer package?
There is `GetLoader` fce that is inside loader package, it might be better to have it in the same way.

dustymabe: yeah - looks like this can't be done because then there would be an import cycle because `transformer.go` would need to import `transformer/kubernetes`, which imports `transformer`. 

```
import cycle not allowed
```

dustymabe:
this is probably a change that needs 2 reviewers. @janetkuo, can you review?

kadel:
> this is probably a change that needs 2 reviewers. @janetkuo, can you review?

Agreed, second pair of eyes would be great.

ngtuna:
LGTM. Thanks @dustymabe :+1:

PR #245, opened 2016-10-27
Add support for user directive

### LINKED ISSUE #244: User directive from docker-compose is siletly ignored

kadel:
This is little a bit user unfriendly :-(

Issue is that in docker-compose you can set `user` by user name or uid. But K8S allows you to set user only by its uid (has to be int).

This is why I added warning that tells user that `user` directive will be ignored if its not a number.
Here is user unfriendly part, in docker-compose even if you set user by it uid, it has to be string. (user: "2324") If you set it as int (user: 2324) libcompose parser will complain about this not being string.

So now Kompose requires `user` to be number but written as string :-(

fixes #244

### DIFF
--- a/pkg/transformer/kubernetes/k8sutils.go
+++ b/pkg/transformer/kubernetes/k8sutils.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -284,12 +285,22 @@ func UpdateKubernetesObjects(name string, service kobject.ServiceConfig, objects
...
+		securityContext := &api.SecurityContext{}
 		if service.Privileged == true {
-			template.Spec.Containers[0].SecurityContext = &api.SecurityContext{
-				Privileged: &service.Privileged,
+			securityContext.Privileged = &service.Privileged
+		}
+		if service.User != "" {
+			uid, err := strconv.ParseInt(service.User, 10, 64)
+			if err != nil {
+				logrus.Warn("Ignoring User directive. User has to specfied as uid (numeric).")
```
cdrage: Just a little spelling error :)

`Ignoring user directive. User to be specified as a UID (numeric).`

pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -284,12 +285,22 @@ func UpdateKubernetesObjects(name string, service kobject.ServiceConfig, objects
...
-		// Configure the container privileged mode
+
+		securityContext := &api.SecurityContext{}
 		if service.Privileged == true {
-			template.Spec.Containers[0].SecurityContext = &api.SecurityContext{
-				Privileged: &service.Privileged,
+			securityContext.Privileged = &service.Privileged
+		}
+		if service.User != "" {
+			uid, err := strconv.ParseInt(service.User, 10, 64)
```
cdrage: What are these numbers? `10` and `64`?

cdrage: Ahhhh, makes sense xD

cdrage:
Failing tests?

kadel:
> Failing tests?
yeah :-( It adds empty security context for every container (`"securityContext": {}`) I have to change that or update tests

kadel:
fixed and rebased

ericchiang:
unit tests?

kadel:
rebased and added  test

ngtuna:
@kadel It needs rebase again

kadel:
rebased

PR #248, opened 2016-10-27
Adding support for choosing empty volumes

### LINKED ISSUE #226: provide easy option for users in setup without PVs
If a user is targeting an env without PVs it would be great if we gave them an easy option for getting up and running. Rather than having them create PVs (which can be complicated or maybe even not allowed if they don't control their environment) I think we should give them the option `--volume=empty`  to substitute in a `emptyDir` for a PVC. The --volume option would could default to `--volume=pvc`. 

This would allow them to get going without having to have it all set up, but would keep sane defaults for people who have real environments with appropriate volume setups. 

More information on `emptyDir` [here](http://kubernetes.io/docs/user-guide/volumes/#emptydir)

dustymabe:
Fixes #226

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -223,6 +223,8 @@ func (k *Kubernetes) ConfigVolumes(name string, service kobject.ServiceConfig) (
 	volumes := []api.Volume{}
 	var PVCs []*api.PersistentVolumeClaim
 
+	useEmptyVolumes := true
+
 	var count int
 	for _, volume := range service.Volumes {
 		volumeName, host, container, mode, err := transformer.ParseVolume(volume)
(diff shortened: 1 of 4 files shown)


dustymabe:
This PR depends on #240 - it is rebased on top of that PR

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -228,41 +234,70 @@ func ConfigVolumes(name string, service kobject.ServiceConfig) ([]api.VolumeMoun
 			logrus.Warningf("Failed to configure container volume: %v", err)
 			continue
 		}
+
+		// check if ro/rw mode is defined, default rw
+		readonly := len(mode) > 0 && mode == "ro"
```
kadel: why `len(mode)>0`?

kadel:
tested and it works great

pkg/transformer/kubernetes/kubernetes.go
dustymabe: This was here before. I just moved it up before the if statement block of code

dustymabe:
@kadel I added two more commits. 7e4c68b is self explanatory. 301b5c2 is because I'd like to be able to run `kompose up --emptyvols`.

Should we reconsider allowing people to pass in the "convert options" for up/down?

ngtuna:
@dustymabe Code LGTM. I will step forward to approve this PR. If @kadel doesn't have any more feedback then I will merge it if you want.

ngtuna:
@kadel Could you merge it? It should be great to have it in the release

PR #254, opened 2016-10-28
Match case with API objects when printing to terminal

dustymabe:


ngtuna:
@janetkuo Could you merge it ? as @dustymabe doesn't have write access.

PR #261, opened 2016-10-31
Modify command in initializing unit tests

cdrage:
This cleans up the current script we have to a simple one-liner to test
unit tests.

### DIFF
--- a/script/test-unit
+++ b/script/test-unit
@@ -3,24 +3,4 @@ set -e
 
 source "$(dirname "$BASH_SOURCE")/.build"
 
-find_dirs() {
-  (
-    find . -not \( \
-        \( \
-          -path './vendor/*' \
-        \) -prune \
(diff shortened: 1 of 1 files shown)


script/test-unit
```diff
@@ -1,26 +1,4 @@
...
-if [ -z "$TEST_DIRS" ]; then
-  TEST_DIRS=$(find_dirs '*_test.go')
-fi
-
-TESTS_FAILED=()
-
-for dir in $TEST_DIRS; do
-  go test "${TEST_FLAGS[@]}" "./${dir}"
-done
+go test -race -cover -v $( go list github.com/kubernetes-incubator/kompose/... | grep -v '/vendor/' )
```
kadel: This looks much cleaner. To be honest I don't know why it was done using `find`. All other projects that I've looked at are also using `find`.
Lets try this and see what it will do ;-)

I would do just one change. To make sure same flags are used for tests and build. Can you add `BUILD_FLAGS` from `.build`?

cdrage:
@kadel updated!

Issue #247, opened 2016-10-27
Why do tests take so long to run?


cdrage:
I'm trying to figure out why the tests take so long to run, even though we only have one test file: `kubernetes_test.go`.

When I run `htop` during the test build, it's showing a lot of vendor tests being ran, is this normal?

sebgoa:
yes that is a big one. I noticed the same thing.

We need to fix this asap.

ericchiang:
> When I run htop during the test build, it's showing a lot of vendor tests being ran, is this normal?

`go test ./...` will run test files in the `vendor` directory. Are you not filtering those packages out properly?

kadel:
Times on my laptop:

go test arguments     |  duration
---------------------------|-------------
`-cover -race `|   208.25s 
` -cover `         |    14.48s  
no args                     | 14.35s

I've run it like this
`time go test  -v $( go list github.com/kubernetes-incubator/kompose/... | grep -v '/vendor/' )`


So it looks like race condition detector is to blame. 
I don't think that we need it in automated tests.

cdrage:
@kadel 

Create a PR for this :) #263

ericchiang:
@kadel that's because you're not caching the build assets and need to rebuild every time.

```
$ pkgs=$( go list github.com/kubernetes-incubator/kompose/... | grep -v '/vendor/' )
$ # install test dependencies
$ go test -v -i -race $pkgs
$ time go test -race $pkgs
?   	github.com/kubernetes-incubator/kompose	[no test files]
ok  	github.com/kubernetes-incubator/kompose/cli/app	1.062s
?   	github.com/kubernetes-incubator/kompose/cli/command	[no test files]
?   	github.com/kubernetes-incubator/kompose/pkg/kobject	[no test files]
?   	github.com/kubernetes-incubator/kompose/pkg/loader	[no test files]
?   	github.com/kubernetes-incubator/kompose/pkg/loader/bundle	[no test files]
?   	github.com/kubernetes-incubator/kompose/pkg/loader/compose	[no test files]
?   	github.com/kubernetes-incubator/kompose/pkg/transformer	[no test files]
ok  	github.com/kubernetes-incubator/kompose/pkg/transformer/kubernetes	1.070s
?   	github.com/kubernetes-incubator/kompose/pkg/transformer/openshift	[no test files]
?   	github.com/kubernetes-incubator/kompose/version	[no test files]

real	0m6.592s
user	0m15.668s
[...]

PR #263, opened 2016-11-01
Improve test times by removing "-race"

cdrage:
Improve test times by removing the -race condition on running the unit
tests


ericchiang:
Please see my comment here https://github.com/kubernetes-incubator/kompose/issues/247#issuecomment-257672539

The race detector isn't your issue, its the fact the unit test script doesn't cache builds combined with kompose having a _huge_ number of dependencies.

PR #278, opened 2016-11-09
Add initial support for bringing up a kubernetes cluster using miniku…

cab105:
…be #156


ericchiang:
Per my comment here (https://github.com/kubernetes-incubator/kompose/issues/156#issuecomment-259847395) I think that this doesn't belong in kompose. I'll let an owner have the finial say though.

Issue #281, opened 2016-11-15
Add coveralls


sebgoa:
to track our unit test coverage, we should setup coveralls.io, it is used in the main repo.

then we need to work on the unit-tests

cdrage:
@sebgoa someone will have to login to coveralls.io to enable it, but I've created #287

sebgoa:
@sarahnovotny sorry to ping you directly, but do you know how we can enable `coveralls` for kompose. I am assuming there is a kubernetes project account.

sebgoa:
In the meantime, I logged into coveralls and requested access for kubernetes-incubator, no clue to whom the request went. Will see.

cdrage:
@sebgoa @sarahnovotny thank you both. Yeah, the PR (https://github.com/kubernetes-incubator/kompose/pull/287) failed due to no access (timeout to coveralls)

PR #287, opened 2016-11-16
Update TRAVIS CI to add coveralls

cdrage:
Updates travis to enable coveralls support.

Fixes https://github.com/kubernetes-incubator/kompose/issues/281

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,9 @@ language: go
 go:
   - 1.6
 
+before_install:
+  - go get github.com/mattn/goveralls
+
 install:
   - true
 
@@ -23,3 +26,4 @@ script:
   - cp kompose $GOPATH/bin/
 
   - make test-cmd
+  - goveralls -service=travis-ci
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Kompose (Kubernetes + Compose)
 
 [![Build Status](https://travis-ci.org/kubernetes-incubator/kompose.svg?branch=master)](https://travis-ci.org/kubernetes-incubator/kompose)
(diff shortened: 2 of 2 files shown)


sebgoa:
I have access to coveralls via incubator. I will test this today.

cdrage:
@sebgoa done. now let's see if these tests pass :)

cdrage:
Currently this isn't working for some odd-reason (timing out). I'll be looking at an alternative fix.

sebgoa:
ok, I am going to merge it and fix it

kadel:
Why merge it if its not working?
It looks like it is missing token https://coveralls.zendesk.com/hc/en-us/articles/201342809-Go

sebgoa:
because I was tired of seeing this PR pending and seeing our code coverage on the front page will be good motivation to improve unit tests.

afaik, you don't need a token for public repos...

but yeah it is failing.

kadel:
To be honestly I don't see reason why it couldn't be fixed in this PR.

Now front page shows no coverage and failed tests.

I've done some digging around, and it is timeouting because it runs all unit test again for every package as separate `go test` run.
like this:
```
for package in all_packages:
   go test package
```
every go test package does new build, so it takes really long time, and travis-ci jobs has 15min timeout.

Root cause for this is that for some reason `go test ` can't collect coverage profile for multiple packages.
So you have to run it for every package again :-(


We could run tests and collect all coverageprofiles into one file  and than call `goveralls` with `-coverprofile` to only submit results to coveralls.io  as described here https://gist.github.com/rjeczalik/6f01430e8554bf59b88e

But it still requires to run `go test` for every package separately.

kadel:
this is how kubernetes is dealing with this https://github.com/kubernetes/kubernetes/blob/master/hack/make-rules/test.sh#L228

PR #329, opened 2016-12-07
Report code coverage to coveralls

kadel:
quick fix for Coveralls problem 
#287

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,17 @@ go:
 
 before_install:
   - go get github.com/mattn/goveralls
+  - go get github.com/modocache/gover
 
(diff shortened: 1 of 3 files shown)


sebgoa:
LGTM

kadel:
lets wait untill travis job finishes.

I only tested this on my fork.

kadel:
Wheeee, it works ;-)

Only down side is that number it reports is  coverage of lines from files that have tests.
So packages without any tests are not included in this.

kadel:
This was quickest solution that I could think of. 
It would be great to get something little bit cleaner.

PR #363, opened 2017-01-05
add deploy/undeploy pod only

ngtuna:
ref: #342

Ping @surajssd @kadel

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -583,6 +583,12 @@ func (k *Kubernetes) Deploy(komposeObject kobject.KomposeObject, opt kobject.Con
 				return err
 			}
 			logrus.Infof("Successfully created Ingress: %s", t.Name)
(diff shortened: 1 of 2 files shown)


ngtuna:
@surajssd @kadel Do you have idea why my PR pulling down the coveralls ? :-(

kadel:
@ngtuna because you added new lines that are not covered by tests :-)

ngtuna:
@kadel oh now I got it. Test is coming up.

kadel:
@ngtuna I'm not sure if this can be currently tested properly without mocking kubernetes :-( or starting cluster. 
But it would be great if you have any idea how to do it.
Otherwise I wouldn't worry about in this PR. We will have to figure out a way how to test up for everything later.

ngtuna:
@kadel Right. I made this commit last night but then came up with no idea how to add a test. Will investigate it more late today.

ngtuna:
@kadel I added commit for OpenShift

PR #373, opened 2017-01-11
error out if controller object is specified with "restart: on-failure"

### LINKED ISSUE #354: error should be displayed If controller object is specified  and `restart: on-failure`
```yaml
version: "2"

services:
  foo:
    image: "busybox"
    restart: "on-failure"
    command: ["sleep", "30s"]
```

```
./kompose convert --deployment
INFO[0000] file "foo-pod.json" created
```

I would expect error or at least warning.

procrypt:
Fixes #354 
CC: @kadel @surajssd


pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -463,6 +463,10 @@ func (k *Kubernetes) Transform(komposeObject kobject.KomposeObject, opt kobject.
 		service := komposeObject.ServiceConfigs[name]
 		var objects []runtime.Object
 
+		// Show error if no controller object is specified and restart: "on-failure"
+		if opt.CreateD && service.Restart == "on-failure" || opt.CreateDS && service.Restart == "on-failure" || opt.CreateDeploymentConfig && service.Restart == "on-failure" || opt.CreateRC && service.Restart == "on-failure" {
```
kadel: this doesn't look right :-(
It should be also handled somewhere else. Don't forget about OpenShift transformer ;-)

script/test/cmd/tests.sh
```diff
@@ -111,7 +111,7 @@ convert::expect_success_and_warning "kompose --bundle $KOMPOSE_ROOT/script/test/
 # Test related to restart options in docker-compose
 # kubernetes test
 convert::expect_success "kompose -f $KOMPOSE_ROOT/script/test/fixtures/restart-options/docker-compose-restart-no.yml convert --stdout -j" "$KOMPOSE_ROOT/script/test/fixtures/restart-options/output-k8s-restart-no.json"
-convert::expect_success "kompose -f $KOMPOSE_ROOT/script/test/fixtures/restart-options/docker-compose-restart-onfail.yml convert --stdout -j" "$KOMPOSE_ROOT/script/test/fixtures/restart-options/output-k8s-restart-onfail.json"
```
kadel: I remove this test? It is valid

pkg/transformer/kubernetes/kubernetes.go
cdrage: k8sutils.go would be a better area. similar to your work on cpu_shares https://github.com/kubernetes-incubator/kompose/pull/294/files

script/test/cmd/tests.sh
surajssd: new failing tests should be added, which will check if it fails on adding more objects on convert.

procrypt:
@kadel @surajssd please review.

cmd/convert.go
```diff
@@ -55,21 +55,25 @@ var convertCmd = &cobra.Command{
...
+			CreateDS:                    ConvertDaemonSet,
+			CreateRC:                    ConvertReplicationController,
+			BuildRepo:                   ConvertBuildRepo,
+			BuildBranch:                 ConvertBuildBranch,
+			CreateDeploymentConfig:      ConvertDeploymentConfig,
+			EmptyVols:                   ConvertEmptyVols,
+			IsDeploymentFlag:            cmd.Flags().Lookup("deployment").Changed,
+			IsDaemonSetFlag:             cmd.Flags().Lookup("daemon-set").Changed,
+			IsReplicationControllerFlag: cmd.Flags().Lookup("replication-controller").Changed,
+			IsDeploymentConfigFlag:      cmd.Flags().Lookup("deployment-config").Changed,
```
cdrage: :+1: on all above :)

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -465,6 +465,10 @@ func (k *Kubernetes) Transform(komposeObject kobject.KomposeObject, opt kobject.
 
 		// Generate pod only and nothing more
 		if service.Restart == "no" || service.Restart == "on-failure" {
+			// Error out if Controller Object is specified with restart: 'on-failure'
+			if opt.IsDeploymentFlag || opt.IsDaemonSetFlag || opt.IsReplicationControllerFlag {
+				logrus.Fatalf("Controller object cannot be specified with restart: 'on-failure'")
+			}
```
cdrage: Don't really understand this part. If a user __doesn't__ pass in these flags, it won't error out?

pkg/transformer/openshift/openshift.go
```diff
@@ -321,6 +321,10 @@ func (o *OpenShift) Transform(komposeObject kobject.KomposeObject, opt kobject.C
 
 		// Generate pod only and nothing more
 		if service.Restart == "no" || service.Restart == "on-failure" {
+			// Error out if Controller Object is specified with restart: 'on-failure'
+			if opt.IsDeploymentConfigFlag {
+				logrus.Fatalf("Controller object cannot be specified with restart: 'on-failure'")
+			}
```
cdrage: same here.. What happens if the user *doesnt* pass in the flag but still supplies an "on-failure"?

procrypt: @cdrage If the user don't pass the flag and still supplies `on-failure`, then `DeploymentConfig` will be created.

cdrage:
Thanks for the comments. This LGTM :) All green from me.

procrypt:
Image

cdrage:
@procrypt Thanks for the awesome work! Merging away :)

PR #383, opened 2017-01-18
Update logging for logrus

cdrage:
This commit disables the timestamp when outputting logs in order to be
more clean / concise.

One of the reasons being that the kompose up / convert / down commands
are *too fast* and thus output's all [0000].

For example, the output will look like this:
INFO I'm a little teapot

Instead of:
INFO[0000] I'm a little teapot


cdrage:
This also updates vendoring in order to import https://github.com/sirupsen/logrus/pull/400

PR #384, opened 2017-01-19
Add Kompose site

cdrage:


kadel:
@cdrage  great! :wink: 

But that giant button is bellow :arrow_down: :rofl:

PR #385, opened 2017-01-20
Container for running tests and Makefile cleanup

kadel:
This adds Dockerfile that can be used to run all Kompose tests locally.

`make test` will run exactly the same tests as travis-ci but running locally in docker image.
This can be used to tests everything before commiting or submitting PR.
Image is automatically  build in docker hub - https://hub.docker.com/r/kompose/tests/ or it can be build locally using `make test-image`.

Automatic build at docker hub is NOT triggered automatically on push. Only time when rebuild is needed is when Dockerfile is changed.

This PR also includes Makefile and `/script/` cleanup.
I've removed all the shell scripts that were just one single command and put those commands directly to Makefile. It is much more easier to figure out what is doing what.


script/test_in_container/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# This Dockefile creates image where all Kompose tests can be run
+
+FROM registry.centos.org/centos/centos:7
```
cdrage: i think we should use the docker hub centos base images

script/test_in_container/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# This Dockefile creates image where all Kompose tests can be run
```
cdrage: missing license at the top

script/test_in_container/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+FROM registry.centos.org/centos/centos:7
+MAINTAINER Tomas Kral <tkral@redhat.com>
```
cdrage: MAINTAINER has been deprecated in 1.13 of Docker :(.

Better to use `LABEL maintainer Tomas Kral <tkral@redhat.com>` or remove it all-together

cdrage:
Just a note though, I feel as though these tests run a lot slower than the original ones? I think it's because there's no cache being passed through?

The part that's *reallyyyy* taking it's time is here:
```
# First install packages that are dependencies of the test. 
go test -i -race -cover ./cmd/... ./pkg/... .
```

Another being that what if I wanted to *just* run cmd or *just* run integration?

script/test_in_container/Dockerfile
kadel: Oh, I didn't know that.
 I'll remove it, I don't think that it is something that we need for image. It was just something that I was used to doing.

kadel:
> Just a note though, I feel as though these tests run a lot slower than the original ones? I think it's because there's no cache being passed through?

Yes, no cache being passed through. 
I'll look at it in separate PR.

> 
> The part that's reallyyyy taking it's time is here:
> 
> # First install packages that are dependencies of the test. 
> go test -i -race -cover ./cmd/... ./pkg/... .

This pre-compiles all the dependencies for test. So it can take some time, but it will speed up running all the `go test`.  
Calculating coverage  for whole project requires running `go test` separately for every package. If you do it without running `go test -i` first it takes really long time (20+mins) :-D.

> Another being that what if I wanted to just run cmd or just run integration?

Yes, I was thinking about it. I would like to do it in separate PR.

I have some other ideas how to speed this up. But than it will run slightly different commands and tests from what is run in travis-ci. I wanted to replicated same thing that travis is doing first, and than add options to run faster but not complete test later.

script/test_in_container/Dockerfile
kadel: fixed, thx for noticing

script/test_in_container/Dockerfile
kadel: I was thinking about it. They should be build from same Dockerfile.
I don't know.

Why would you prefer image from docker hub?

cdrage:
I like how you keep the current tests in as well. But perhaps we can do it so that it's "test-container" or something?

Would be easier for using it on travis, etc.

But another thing I thought of was that if it's in a container, it'll have to download the dependencies / update each time? Or passing vendoring in shouldn't cause that issue?

May be high on bandwidth costs for those in other countries.

kadel:
There is `make test-all` that runs all the tests outside container and `make test` that runs it inside container. 
My idea was that "humans" will run tests in container so i wanted to make command shorter, hence `make test` :-) 

So you are suggesting changing it so `make test` runs everything without container and creating  `make test-container` that doesn't what `make test` is doing right now (running it it all inside container)?

cdrage:
@kadel 
Yup. That's what I'm suggesting. `make test` runs it traditionally. `make test-container` runs it in a container.

kadel:
> @kadel
> Yup. That's what I'm suggesting. make test runs it traditionally. make test-container runs it in a container.

done ;-)

There is just one small catch with `make test` and that it expects that you have $GOPATH/bin in $PATH.
Otherwise test-cmd might fail or  run against older binary.

Makefile
```diff
@@ -13,47 +13,86 @@
...
+
+# build docker image that is used for running all test localy
+.PHONY: test-image
+test-image:
+	docker build -t $(TEST_IMAGE) -f script/test_in_container/Dockerfile script/test_in_container/
+	
+# run all test localy in docker image (image can be build by by build-test-image target)
+.PHONY: test-container
+test-container:
+	docker run -v `pwd`:/opt/tmp/kompose:ro -it $(TEST_IMAGE)
```
cdrage: woo

LGTM! :)

PR #386, opened 2017-01-20
Update to use YAML instead of json

cdrage:


cdrage:
ping @surajssd thanks for noticing this :)

PR #387, opened 2017-01-20
Ignores :z or :Z when passed in as a volume string

cdrage:
We're going to ignore :z / :Z for labeling aka SELinux when being passed
in via Docker Compose.

Closes https://github.com/kubernetes-incubator/kompose/issues/176


cdrage:
Not too sure if this was the most-correct way of doing this @kadel but here's a try at it :)

ankon:
Note that the modes can be a comma-separated list: https://docs.docker.com/engine/reference/run/#/volume-shared-filesystems says 
```
-v, --volume=[host-src:]container-dest[:<options>]: Bind mount a volume.
The comma-delimited `options` are [rw|ro], [z|Z],
[[r]shared|[r]slave|[r]private], and [nocopy].
```

Issue #392, opened 2017-01-25
By default, remove protocol: TCP in conversion to artifacts


cdrage:
By default, TCP is selected.

https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_serviceport

We should not include `protocol: tcp` in the resulting yaml / json conversion since it's already set by default.

PR #394, opened 2017-01-26
Removes the TCP output on the Kubernetes / OpenShift artifacts

cdrage:
If TCP is passed in as the protocol, by default, we add TCP as the
protocol within the Kubernetes or OpenShift artifacts.

By default, TCP is already selected within Kubernetes and thus having
the TCP output is redundant.

This commit checks to see if TCP has already been selected, and if it
has, ignores adding it to the list of ports.

Closes https://github.com/kubernetes-incubator/kompose/issues/392


cdrage:
Tests are gunna fail.

Ignore tests for now (I'll need to fix them as they test if TCP is included). Will work on this.

cdrage:
Tests pass now! :+1: Had to go through _each and every_ JSON file to update it to not include TCP...

Ready for review!

cdrage:
Ping @surajssd @containscafeine @procrypt @kadel @rtnpro for review!

PR #418, opened 2017-02-09
Update Makefile - lazy set for PKGS variable

kadel:
Using lazy set we can run targets that don't $PKGS without requiring glide.

Lazy set is expended when variable is used, not when declared.


cdrage:
LGTM after merge conflicts are resolved :+1:

kadel:
rebased

PR #434, opened 2017-02-21
Abstract out api.pod spec

procrypt:
Fixes: #348 
cc: @containscafeine @kadel @cdrage @surajssd

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -104,6 +104,19 @@ func (k *Kubernetes) CheckUnsupportedKey(komposeObject *kobject.KomposeObject, u
 	return keysFound
 }
 
+// InitPodSpec creates the pod specification
+func (k *Kubernetes) InitPodSpec(name string, image string) api.PodSpec {
+	pod := api.PodSpec{
+		Containers: []api.Container{
+			{
+				Name:  name,
+				Image: image,
+			},
+		},
+	}
+	return pod
+}
+
 // InitRC initializes Kubernetes ReplicationController object
 func (k *Kubernetes) InitRC(name string, service kobject.ServiceConfig, replicas int) *api.ReplicationController {
 	rc := &api.ReplicationController{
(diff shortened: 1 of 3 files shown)


kadel:
PodSpec is also used in OpenShift transformer. It would be great to use podspec function also there.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -104,6 +104,19 @@ func (k *Kubernetes) CheckUnsupportedKey(komposeObject *kobject.KomposeObject, u
 	return keysFound
 }
 
+// podspec creates the pod specification
+func (k *Kubernetes) podspec(name string, service kobject.ServiceConfig) api.PodSpec {
```
cdrage: should be podSpec (Go name conventions)

kadel: and even better might be initPodSpec similar to other init functions ;-)

procrypt:
@kadel We can't use `PodSpec` function in OpenShift because `PodSpec` in OpenShift is different.
PodSpec in `Kubernetes`
```go
	pod := api.PodSpec{
		Containers: []api.Container{
			{
				Name:  name,
				Image: service.Image,
			},
		},
	}
```

PodSpec in `OpenShift`
```go
	pod := api.PodSpec{
		Containers: []api.Container{
			{
				Name:  name,
				Image: " ",
			},
		},
	}
```

kadel:
I think that if we are abstracting podSpec in one place it would be great to have everywhere else.
 
If you change your `initPodSpec` function it can be used also in OpenShift. 
You don't have to pass whole `service kobject.ServiceConfig` struct down to `initPodSpec`.
All you need is `name` and `image`. It could look like this:

```go
func (k *Kubernetes) InitPodSpec(name string, image string) api.PodSpec {
... 
```

procrypt:
@kadel PR updated.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -104,6 +104,19 @@ func (k *Kubernetes) CheckUnsupportedKey(komposeObject *kobject.KomposeObject, u
 	return keysFound
 }
 
+// "InitPodSpec creates the pod specification"
```
kadel: remove `"` and it will be ok ;-)

procrypt: @kadel tests are passing locally when I run `make test`, no idea what is going wrong in Travis.

cdrage:
LGTM

PR #448, opened 2017-02-23
Add conversion documentation

cdrage:
This adds a document regarding all the values which are converted from
Docker Compose to Kubernetes / OpenShift. Adding both notes as well as
the value that docker-compose is mapped to in Kubernetes / OpenShift.

### DIFF
--- a/docs/conversion.md
+++ b/docs/conversion.md
@@ -1,0 +1,61 @@
+This document outlines all the conversion details regarding `docker-compose.yaml` values to Kubernetes / OpenShift artifacts.
+
+| Value | Version | Supported | K8s / OpenShift | Notes |
+|-------------------|---------|-----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|
+| __SERVICE__ |  |  |  |  |
+| build |  | Y | OpenShift: [BuildConfig](https://docs.openshift.com/enterprise/3.1/dev_guide/builds.html#defining-a-buildconfig) | Converts, but local builds are not yet supported. See issue [97](https://github.com/kubernetes-incubator/kompose/issues/97) |
+| cap_add, cap_drop |  | N |  |  |
+| command |  | Y | [Pod.Spec.Container.Command](https://kubernetes.io/docs/api-reference/v1/definitions/#_v1_container) |  |
+| cgroup_parent |  | N |  | No compatibility with Kubernetes / OpenShift. Limited use-cases with Docker. |
+| container_name |  | Y | Mapped to both [Metadata.Name](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/identifiers.md) and [Deployment.Spec.Containers.Name](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/identifiers.md) |  |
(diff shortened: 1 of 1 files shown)


docs/conversion.md
```diff
@@ -0,0 +1,64 @@
...
+| Service           |           |             |                                                                                                                                                                                                                                                                          |                                                                                                                             |
+| build             |           | YES         | [BuildConfig](https://docs.openshift.com/enterprise/3.1/dev_guide/builds.html#defining-a-buildconfig)                                                                                                                                                                    | Converts, but local builds are not yet supported. See issue [97](https://github.com/kubernetes-incubator/kompose/issues/97) |
```
kadel: Maybe mention that build is supported only for OpenShift.
For Kubernetes `build` in unsupported

docs/conversion.md
```diff
@@ -0,0 +1,64 @@
...
+| sysctls           |           | UNSUPPORTED |                                                                                                                                                                                                                                                                          |                                                                                                                             |
+| ulimits           |           | UNSUPPORTED |                                                                                                                                                                                                                                                                          |                                                                                                                             |
```
kadel: Kubernets issue about ulimit support - https://github.com/kubernetes/kubernetes/issues/3595

docs/conversion.md
```diff
@@ -0,0 +1,64 @@
...
+| build             |           | YES         | [BuildConfig](https://docs.openshift.com/enterprise/3.1/dev_guide/builds.html#defining-a-buildconfig)                                                                                                                                                                    | Converts, but local builds are not yet supported. See issue [97](https://github.com/kubernetes-incubator/kompose/issues/97) |
+| cap_add, cap_drop |           | YES         | N/A                                                                                                                                                                                                                                                                      |                                                                                                                             |
```
kadel: I think there is a mistake in the code. We are reading cap_add and cap_drop but we are not using it anywhere.

PR https://github.com/kubernetes-incubator/kompose/pull/468 is adding `cap_add` and `cap_drop` to  unsupported keys.

docs/conversion.md
```diff
@@ -0,0 +1,64 @@
...
+| devices           |           | UNSUPPORTED | N/A                                                                                                                                                                                                                                                                      | Not supported within Kubernetes, see this [issue](https://github.com/kubernetes/kubernetes/issues/5607)                     |
+| depends_on        |           | UNSUPPORTED | N/A                                                                                                                                                                                                                                                                      | Elaborate on why this isn't supported and what happens when it's converted. Multiple users in #kompose has voiced concerns  |
```
kadel: What concerns were voiced?
We have `depends_on` in unsupported keys. I'm not sure if that is right, maybe yes, maybe no :-D
It needs more explanation.
`depends_on` defines order in which containers should start. Problem is that there is no way to ensure that in Kubernetes. 
Questionis is if it matters or not :-)

kadel:
I should also have link to this document in readme

cdrage:
@kadel I agree. I'll add an "index" to the README eventually.

I've updated the PR based on your comments as well as converted it to a "compact" table instead of having all the spacing everywhere.

cdrage:
@kadel Updated again, I shortened the unsupported column to Y/N as well as labeled the separate sections by boldness (volumes, services, networks, etc.). Looks better now :+1:

PR #449, opened 2017-02-24
Test with multiple go versions

kadel:
`go vet` in go 1.8 started complaining about tag format.

```
pkg/kobject/kobject.go:60: struct field tag `compose:"image",bundle:"Image"` not compatible with reflect.StructTag.Get: key:"value" pairs not separated by spaces
```

This fixes tag formatting in KObject and updates travis.yaml so it runs tests on multiple go versions 1.6, 1.7 and 1.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,14 @@ sudo: required
 
 language: go
 
+# make this also work for forks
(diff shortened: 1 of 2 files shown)


cdrage:
LGTM :+1:

PR #451, opened 2017-02-24
Update the release script again :)

cdrage:


script/release.sh
```diff
@@ -81,11 +81,7 @@ replaceversion() {
   echo "1. Replaced version in version.go"
   sed -i "s/$1/$2/g" cmd/version.go
 
-  echo "2. Replace GITCOMMIT value in version.go"
-  VERSION=`git log -1 --pretty=format:%h`
-  sed -i "s|.*GITCOMMIT = .*|\tGITCOMMIT = \"$VERSION\"|" cmd/version.go
```
kadel: why this?, we don't need to do that no?

kadel: GITCOMMIT is modified during build - https://github.com/kubernetes-incubator/kompose/blob/master/Makefile#L18

cdrage: no we don't have to do that, that's why i removed it.

kadel: ahhh, my BAD :blush: 

I didn't realized that you removed it, I was reading it as you add it :-D 

Thank god it is already  FRIDAY :-D

Issue #439, opened 2017-02-22
Error / Support for: links


cdrage:
No error / warning when using this key.

kadel:
`links` can be silently ignored.
Links are just injecting env variables with information about ips and ports. Kubernetes is injecting same env variables to pods by default.

cdrage:
:+1:

PR #459, opened 2017-03-01
unsupported keys

procrypt:
@cdrage added `sysctls` to unsupported keys since we didn't find anything akin in Kubernetes.
Added `links` to unsupported keys since containers in the same pod can find and communicate with each other using `localhost:containerPort`

Fixes: #441 #439


kadel:
I don't think that we should display warning for `links`.
Only think `link` does is create env variables, same variables are create by Kubernetes for every service.

We can safely and silently ignore it.

cdrage:
Agreed with @kadel and thanks for adding sysctls to ignore list :)

Issue #323, opened 2016-12-05
Functional tests for OpenShift down


procrypt:
Need functional tests for #280

ericchiang:
Shouldn't this be addressed in the PR? Why the extra issue?

kadel:
Almost all code in this touches OpenShift cluster.
To test this, it will require setting up OpenShift cluster or mocking its api.

Same situation is with up/down for k8s. We will have to figure out how we will test all this.

We don't want to block PR on that. So we created this separate issue.

surajnarwade:
@cdrage @kadel, we can close this issue as functional tests are being added for `kompose down` by @ashetty1

PR #460, opened 2017-03-01
Adding kompose up/down tests for openshift

ashetty1:
Adding kompose up/down tests for openshift as part of #323.

The script runs kompose up/down using etherpad docker-compose file with `oc cluster up`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,12 @@ test-unit-cover:
 	# generate go test commands using go list and run go test for every package separately 
 	go list -f '"go test -race -cover -v -coverprofile={{.Dir}}/.coverprofile {{.ImportPath}}"' github.com/kubernetes-incubator/kompose/...  | grep -v "vendor" | xargs -L 1 -P4 sh -c 
 
+
+# run openshift up/down tests
+.PHONY: test-openshift
+test-openshift:
+	./script/test_in_openshift/run.sh
+
 # run commandline tests
 .PHONY: test-cmd
 test-cmd:
@@ -89,7 +95,7 @@ test: check-vendor validate test-unit-cover install test-cmd
 .PHONY: test-image
 test-image:
 	docker build -t $(TEST_IMAGE) -f script/test_in_container/Dockerfile script/test_in_container/
-	
+
 # run all test localy in docker image (image can be build by by build-test-image target)
 .PHONY: test-container
 test-container:
--- a/script/test_in_openshift/lib.sh
+++ b/script/test_in_openshift/lib.sh
@@ -1,0 +1,217 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+function convert::print_msg () {
+    echo ""
+    tput setaf 4
+    tput bold
+    echo -e "$@"
+    tput sgr0
+    echo ""
+}
+
+function install_oc_client () {
+    # Valid only for Travis
+    convert::print_msg "Installing oc client binary ..."
+    sudo sed -i 's:DOCKER_OPTS=":DOCKER_OPTS="--insecure-registry 172.30.0.0/16 :g' /etc/default/docker
+    sudo mv /bin/findmnt /bin/findmnt.backup
+    sudo /etc/init.d/docker restart
+    # FIXME
+    wget https://github.com/openshift/origin/releases/download/v1.5.0/openshift-origin-client-tools-v1.5.0-031cbe4-linux-64bit.tar.gz -O /tmp/oc.tar.gz 2> /dev/null > /dev/null
+    mkdir /tmp/ocdir && cd /tmp/ocdir && tar -xvvf /tmp/oc.tar.gz > /dev/null
+    sudo mv /tmp/ocdir/*/oc /usr/bin/
+}
+
+
+function convert::oc_cluster_up () {
+
+    oc cluster up; exit_status=$?
+
+    if [ $exit_status -ne 0 ]; then
+	FAIL_MSGS=$FAIL_MSGS"exit status: $exit_status\n";
+	convert::print_fail "oc cluster up failed.\n"
+	exit $exit_status
+    fi
+
+    convert::run_cmd "oc login -u system:admin"
+}
+
+function convert::oc_cluster_down () {
+
+    convert::run_cmd "oc cluster down"
+    exit_status=$?
+
+    if [ $exit_status -ne 0 ]; then
+	FAIL_MSGS=$FAIL_MSGS"exit status: $exit_status\n"
+	exit $exit_status
+    fi
+
+}
+
+function convert::kompose_up_check () {
+    # A function to check if the pods are up in 'Running' state
+    # WARNING: This function has a limitation that it works only for 2 pods.
+    # TODO: Make this function work for any no. of pods
+    # Usage: -p for pod name, -r replica count
+    local retry_up=0
+
+    while getopts ":p:r:" opt; do
+	case $opt in
+	    p ) pod=$OPTARG;;
+	    r ) replica=$OPTARG;;
+	esac
+    done
+
+    if [ -z $replica ]; then
+       replica_1=1
+       replica_2=1
+    else
+       replica_1=$replica
+       replica_2=$replica
+    fi
+
+    pod_1=$( echo $pod | awk '{ print $1 }')
+    pod_2=$( echo $pod | awk '{ print $2 }')
+
+    query_1='grep ${pod_1} | grep -v deploy'
+    query_2='grep ${pod_2} | grep -v deploy'
+
+    query_1_status='Running'
+    query_2_status='Running'
+
+    is_buildconfig=$(oc get builds --no-headers | wc -l)
+
+    if [ $is_buildconfig -gt 0 ]; then
+	query_1='grep ${pod_1} | grep -v deploy | grep -v build'
+	query_2='grep build | grep -v deploy'
+	query_2_status='Completed'
+	replica_2=1
+    fi
+
+    convert::print_msg "Waiting for the pods to come up ..."
+
+    # FIXME: Make this generic to cover all cases
+    while [ $(oc get pods | eval ${query_1} | awk '{ print $3 }' | \
+		     grep ${query_1_status} | wc -l) -ne $replica_1 ] ||
+	      [ $(oc get pods | eval ${query_2} | awk '{ print $3 }' | \
+			 grep ${query_2_status} | wc -l) -ne $replica_2 ]; do
+
+	if [ $retry_up -lt 240 ]; then
+	    retry_up=$(($retry_up + 1))
+	    sleep 1
+	else
+	    convert::print_fail "kompose up has failed to bring the pods up\n"
+	    oc get pods
+	    exit 1
+	fi
+
+    done
+
+    # Wait
+    sleep 2
+
+    # If pods are up, print a success message
+    if [ $(oc get pods | eval ${query_1} | awk '{ print $3 }' | \
+		  grep ${query_1_status} | wc -l) -eq $replica_1 ] &&
+	   [ $(oc get pods | eval ${query_2} | awk '{ print $3 }' | \
+		      grep ${query_2_status} | wc -l) -eq $replica_2 ]; then
+	oc get pods
+	convert::print_pass "All pods are Running now. kompose up is successful."
+    fi
+}
+
+function convert::kompose_down_check () {
+    # Checks if the pods have been successfully deleted
+    # with 'kompose down'.
+    # Usage: convert::kompose_down_check <num_of_pods>
+
+    local retry_down=0
+    local pod_count=$1
+
+    convert::print_msg "Waiting for the pods to go down ..."
+
+    while [ $(oc get pods | wc -l ) != 0 ] &&
+	  [ $(oc get pods | grep -v deploy | grep 'Terminating' | wc -l ) != $pod_count ]; do
+	if [ $retry_down -lt 120 ]; then
+	    retry_down=$(($retry_down + 1))
+	    sleep 1
+	else
+	    convert::print_fail "kompose down has failed\n"
+	    oc get pods
+	    exit 1
+	fi
+    done
+
+    # Wait
+    sleep 2
+
+    # Print a message if all the pods are down
+    if [ $(oc get pods | wc -l ) == 0 ] ||
+       [ $(oc get pods | grep -v deploy | grep 'Terminating' | wc -l ) == $pod_count ]; then
+	convert::print_pass "All pods are down now. kompose down successful.\n"
+	oc get pods
+    fi
+}
+
+function convert::oc_cleanup () {
+    oc delete bc,rc,rs,svc,is,dc,deploy,ds,builds,route --all > /dev/null
+}
+
+function convert::oc_check_route () {
+    local route_key=$1
+    if [ $route_key == 'true' ]; then
+	route_key='xip.io'
+    fi
+
+    if [ $(oc get route | grep ${route_key} | wc -l ) -gt 0 ]; then
+	convert::print_pass "Route *.${route_key} has been exposed"
+    else
+	convert::print_fail "Route *.${route_key} has not been exposed\n"
+    fi
+
+    echo ""
+    oc get route
+}
+
+function convert::kompose_up () {
+    # Function for running 'kompose up'
+    # Usage: convert::kompose_up <docker_compose_file>
+    local compose_file=$1
+    convert::print_msg "Running kompose up ..."
+    kompose --provider=openshift --emptyvols -f $compose_file up
+    exit_status=$?
+
+    if [ $exit_status -ne 0 ]; then
+	convert::print_fail "kompose up has failed\n"
+	exit 1
+    fi
+}
+
+
+function convert::kompose_down () {
+    # Function for running 'kompose down'
+    # Usage: convert::kompose_down <docker_compose_file>
+    local compose_file=$1
+    convert::print_msg "Running kompose down ..."
+    kompose --provider=openshift -f $compose_file down
+    exit_status=$?
+
+    if [ $exit_status -ne 0 ]; then
+	convert::print_fail "kompose down has failed\n"
+	exit 1
+    fi
+}
--- a/script/test_in_openshift/run.sh
+++ b/script/test_in_openshift/run.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Test case for kompose up/down with etherpad
+
+KOMPOSE_ROOT=$(readlink -f $(dirname "${BASH_SOURCE}")/../../..)
+source $KOMPOSE_ROOT/kompose/script/test/cmd/lib.sh
+source $KOMPOSE_ROOT/script/test_in_openshift/lib.sh
+openshift_exit_status=0
+
+convert::start_test "Functional tests on OpenShift"
+
+if [[ -n "${TRAVIS}" ]]; then
+    install_oc_client
+fi
+
+if [ -z $(whereis oc | awk '{ print $2 }') ]; then
+    convert::print_fail "Please install the oc binary to run tests\n"
+    exit 1
+fi
+
+convert::oc_cluster_up
+
+for test_case in $KOMPOSE_ROOT/script/test_in_openshift/tests/*; do
+    $test_case; exit_status=$?
+    if [ $exit_status -ne 0 ]; then
+	openshift_exit_status=1
+    fi
+    convert::oc_cleanup
+done
+
+convert::oc_cluster_down
+
+exit $openshift_exit_status
--- a/script/test_in_openshift/tests/buildconfig.sh
+++ b/script/test_in_openshift/tests/buildconfig.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Test case for buildconfig on kompose
+
+KOMPOSE_ROOT=$(readlink -f $(dirname "${BASH_SOURCE}")/../../..)
+source $KOMPOSE_ROOT/script/test/cmd/lib.sh
+source $KOMPOSE_ROOT/script/test_in_openshift/lib.sh
+
+convert::print_msg "Testing buildconfig on kompose"
+
+docker_compose_file="${KOMPOSE_ROOT}/examples/buildconfig/docker-compose.yml"
+
+# Run kompose up
+convert::kompose_up $docker_compose_file
+
+# Check if the pods are up.
+convert::kompose_up_check -p foo
+
+# Kompose down for buildconfig fails being tracked at #382
+# convert::kompose_down $docker_compose_file
+
+# convert::kompose_down_check 2
--- a/script/test_in_openshift/tests/redis-replica-2.sh
+++ b/script/test_in_openshift/tests/redis-replica-2.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Test case for checking replicas option with kompose
+
+KOMPOSE_ROOT=$(readlink -f $(dirname "${BASH_SOURCE}")/../../..)
+source $KOMPOSE_ROOT/script/test/cmd/lib.sh
+source $KOMPOSE_ROOT/script/test_in_openshift/lib.sh
+
+convert::print_msg "Running tests for replica option"
+
+# Run kompose up
+kompose --provider=openshift --emptyvols --replicas 2 -f ${KOMPOSE_ROOT}/examples/docker-compose-counter.yaml up; exit_status=$?
+
+if [ $exit_status -ne 0 ]; then
+    convert::print_fail "kompose up has failed"
+    exit 1
+fi
+
+
+# Check if redis and web pods are up. Replica count: 2
+convert::kompose_up_check -p "redis web" -r 2
+
+# Run Kompose down
+convert::kompose_down ${KOMPOSE_ROOT}/examples/docker-compose-counter.yaml
+
+convert::kompose_down_check 4
--- a/examples/buildconfig/build/Dockerfile
+++ b/examples/buildconfig/build/Dockerfile
@@ -1,5 +1,4 @@
-FROM busybox
-
+FROM busybox:1.26.2
 
 RUN touch /test
 
--- a/script/test_in_openshift/compose-files/docker-compose-command.yml
+++ b/script/test_in_openshift/compose-files/docker-compose-command.yml
@@ -1,0 +1,10 @@
+version: '2'
+
+services:
+  base1:
+    image: busybox
+    command: ['sleep','10000']
+
+  base2:
+    image: busybox
+    entrypoint: ['sleep','10000']
--- a/script/test_in_openshift/tests/entrypoint-command.sh
+++ b/script/test_in_openshift/tests/entrypoint-command.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+KOMPOSE_ROOT=$(readlink -f $(dirname "${BASH_SOURCE}")/../../..)
+source $KOMPOSE_ROOT/script/test/cmd/lib.sh
+source $KOMPOSE_ROOT/script/test_in_openshift/lib.sh
+
+convert::print_msg "Running tests with entrypoint/command option"
+
+docker_compose_file="${KOMPOSE_ROOT}/script/test_in_openshift/compose-files/docker-compose-command.yml"
+
+# Run kompose up
+convert::kompose_up $docker_compose_file
+
+convert::kompose_up_check -p 'base1 base2'
+
+# Run Kompose down
+convert::kompose_down $docker_compose_file
+
+convert::kompose_down_check 2
+
--- a/script/test_in_openshift/tests/etherpad.sh
+++ b/script/test_in_openshift/tests/etherpad.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Test case for kompose up/down with etherpad
+
+KOMPOSE_ROOT=$(readlink -f $(dirname "${BASH_SOURCE}")/../../..)
+source $KOMPOSE_ROOT/script/test/cmd/lib.sh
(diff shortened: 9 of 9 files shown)


Makefile
```diff
@@ -83,13 +89,13 @@ check-vendor:
 
 # Run all tests
 .PHONY: test
-test: check-vendor validate test-unit-cover install test-cmd
+test: check-vendor validate test-unit-cover install test-cmd test-openshift
```
kadel: I don't think that `test-openshift` should be part of `make test`

script/test_in_openshift/run.sh
```diff
@@ -0,0 +1,114 @@
...
+    exit 1;
+fi
+
+oc login -u system:admin;
+
+# Env variables for etherpad
+export $(cat ${KOMPOSE_ROOT}/kompose/script/test/fixtures/etherpad/envs)
+
+# Run kompose up
+kompose --emptyvols --provider=openshift -f $KOMPOSE_ROOT/kompose/script/test/fixtures/etherpad/docker-compose.yml up; result=$?;
```
kadel: It would be great to separate parts that are running tests from parts that are doing openshift setup.
Tests should be in different files from setup script

kadel:
please add `make test-openshift` to travis.yml (after `make test`) so we can see it in action ;-)

ashetty1:
I signed it!

cdrage:
A lot of the `;` can be removed from the back file

script/test_in_openshift/run.sh
```diff
@@ -0,0 +1,114 @@
...
+
+# Run oc cluster up
+oc cluster up; result=$?;
+
+if [ $result -ne 0 ]; then
+    echo "oc cluster up failed"
+    exit 1;
+fi
+
+oc login -u system:admin;
```
surajssd: we don't need to be system:admin for everything, it's only needed when creating pvc.

ashetty1: Sure, will fix this.

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
```
cdrage: missing license

script/test_in_openshift/tests/buildconfig.sh
```diff
@@ -0,0 +1,78 @@
...
+if [ $exit_status -ne 0 ]; then
+    convert::print_fail "kompose down failed"
+    exit 1;
+fi
+
+sleep 60;
+
+convert::kompose_down_check
+
+if [ $(oc get pods | wc -l ) == 0 ] ; then
```
kadel: Can this be part of `kompose_down_check`? Does it make sense?

script/test_in_openshift/tests/buildconfig.sh
```diff
@@ -0,0 +1,78 @@
...
+   convert::print_fail "kompose up fails"
+   exit 1
+fi
+
+
+# Wait
+sleep 60;
+
+retry_up=0
+while [ "$(oc get pods | grep foo | grep -v deploy | grep -v build | awk '{ print $3 }')" != 'Running'  ] ||
```
kadel: Exactly same block is repeating in other test cases. It might be better to have this as function in `libs.sh`

script/test_in_openshift/tests/buildconfig.sh
```diff
@@ -0,0 +1,78 @@
...
+	convert::print_fail "kompose up has failed to bring the pods up"
+	exit 1
+    fi
+done
+
+# Wait
+sleep 5;
+
+# Check if all the pods are up
+if [ "$(oc get pods | grep foo | grep -v deploy | grep -v build | awk '{ print $3 }')" == 'Running'  ] &&
```
kadel: Exactly same block is repeating in other test cases. It might be better to have this as function in `libs.sh`

script/test_in_openshift/tests/buildconfig.sh
ashetty1: Sure, will add it as part of the same function

script/test_in_openshift/tests/buildconfig.sh
ashetty1: @kadel Same comment as above: I thought about it, but then I am not sure if the same logic could be used for every case. So wasn't sure about adding it into libs.sh

Let me know if you think otherwise.

script/test_in_openshift/tests/buildconfig.sh
kadel: it should definitely go there, it can have some parameters to make it more generic

script/test_in_openshift/tests/redis-replica-2.sh
```diff
@@ -0,0 +1,80 @@
...
+	exit 1
+    fi
+done
+
+# Wait
+sleep 5;
+
+# Check if all the pods are up
+
+if [ "$(oc get pods | grep redis | grep -v deploy |  awk '{ print $3 }'| grep 'Running' | wc -l)" -eq 2  ] && [ "$(oc get pods | grep web | grep -v deploy | awk '{ print $3 }' | grep 'Running' | wc -l)" -eq 2  ] ; then
```
kadel: this is also candidate for creating function from it and moving it to libs.sh

kadel:
I just noticed that tests take 22mins now :-( 
If I remember correctly there is 30min timeout in travis-ci, so we are getting dangerously close to that :-( 

I guess that `oc cluster up` is taking long as is downloading docker images.

Is possible to cache those images on travis-ci? 
What if we move this test to ci.centos.org. Is it going to be faster? Will it be possible to cache images there?

ashetty1:
@kadel what I could do is not have run `oc cluster up` for every test case. We should have it run it only once.

cdrage:
There's over 7 60 second sleep calls as well as 30 second and 5 second sleep calls all over the place.

@ashetty1 here is some reference code: https://github.com/projectatomic/adb-tests/blob/master/cdrage-atomicapp-ci/functional-tests/providers/openshift.sh

as well as: https://github.com/projectatomic/adb-tests/blob/master/cdrage-atomicapp-ci/functional-tests/

Where we used curl + grep + grepping `oc` and `kubectl` to reduce this time. 

At the moment, this will take wayyyyy too long to run the tests each time. It already takes 10 minutes to run our ordinary tests, but with these tests it'll run even longer.

With the length of the tests, I'd suggest using the CentOS CI instead so we can run both in parallel. I originally suggested *just* Travis CI, but lately Travis has been running slow and it doesn't help that we're adding more and more unit / integration tests each day!

ashetty1:
@cdrage the sleep has been used only to provide some lag while the pods are being brought up. I tried reducing the sleep time to 10, and then modified the kompose_down_check to look for "Terminating' status; with both these changes, the tests took about 12mins. If this looks okay, I shall push the changes her: https://github.com/ashetty1/kompose/blob/entrypoint_cmd/script/test_in_openshift/lib.sh#L157

cdrage:
@ashetty1 
It would be much better if you wrote a bash loop instead of using sleep as it'll really cut down on the deployment times. It's hard to rely just on sleep as we don't know exactly what box the CI test is being ran on at TravisCI and there will be a lot of false positives / having to reset the build.

cdrage:
Look at our solution here: https://github.com/projectatomic/adb-tests/blob/master/cdrage-atomicapp-ci/functional-tests/providers/openshift.sh#L86

ashetty1:
@cdrage Sure, I could get rid of all the sleeping, but then the only issue is about handling failures. How do you think we should handle a scenario where pods aren't getting deleted/terminated?  That's the reason I have the retry count there. Sleep is useful there, I think.

cdrage:
In regards to actually clearing / deleting the pods to run the next test, we didn't run into any of those issues  when deleting / terminating the pods. 

Besides, if a previous test didn't terminate correctly (`kompose down`) then this script is working as intended.

Just have a look at how we ran all the tests here: https://github.com/projectatomic/adb-tests/tree/master/cdrage-atomicapp-ci/functional-tests and you'll see how we did it in terms of terminating / rebuilding the cluster to a clean-slate for each test being ran.

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,183 @@
...
+function install_oc_client () {
+    sudo sed -i 's:DOCKER_OPTS=":DOCKER_OPTS="--insecure-registry 172.30.0.0/16 :g' /etc/default/docker
+    sudo mv /bin/findmnt /bin/findmnt.backup
+    sudo cat /etc/default/docker
+    sudo /etc/init.d/docker restart
+    # FIXME
+    wget https://github.com/openshift/origin/releases/download/v1.4.1/openshift-origin-client-tools-v1.4.1-3f9807a-linux-64bit.tar.gz -O /tmp/oc.tar.gz 2> /dev/null > /dev/null
+    mkdir /tmp/ocdir && cd /tmp/ocdir && tar -xvvf /tmp/oc.tar.gz > /dev/null
+    sudo mv /tmp/ocdir/*/oc /usr/bin/
+}
```
cdrage: I just realized that *all* of this could be a serious issue in terms of what's being ran on the host system..

Shouldn't we error out instead if there's no `oc` on the host system? All of these sudo commands as well as replacing parts of the DOCKER_OPTS (I know that I have custom stuff in there that I don't want overridden / changes)

Obviously for Travis / CentOS CI we'll have this wget magic here, but not for testing.

@kadel do you agree?

script/test_in_openshift/tests/redis-replica-2.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if [ $exit_status -ne 0 ]; then
+    convert::print_fail "Kompose down failed"
+    exit 1
+fi
+
+sleep 60
+
+convert::kompose_down_check
+
```
cdrage: Both of these files are mostly copy-and-paste of all the functions. Is there a way to simplify these into a function?

script/test_in_openshift/run.sh
```diff
@@ -0,0 +1,33 @@
...
+convert::start_test "Functional tests on OpenShift"
+install_oc_client
+
+convert::oc_cluster_up
+for test_case in $KOMPOSE_ROOT/script/test_in_openshift/tests/*; do
+    $test_case
+    sleep 5
+    convert::oc_cleanup
+done
+convert::oc_cluster_down
```
cdrage: :+1:

script/test_in_openshift/lib.sh
ashetty1: @cdrage This is specifically for travis. What I could do is set a env variable "travis=true" in travis.yml, and have that checked before running this function. Does that make sense?

script/test_in_openshift/tests/redis-replica-2.sh
ashetty1: In some sense, yes. But writing a function would mean writing a wrapper for kompose operations. I thought it would be too broad/diverse to handle within a single function.

script/test_in_openshift/lib.sh
cdrage: It does make sense, but people may want to run this test on their host system. I know I'd like to, instead of having to rely on Travis each time. @kadel ?

ashetty1: @cdrage we exit with a message saying they need to have the oc binaries installed?

cdrage: @ashetty1 yes, we remove these sudo commands and mv commands and replace them with a prompt in regards to installing `oc`.

ashetty1: @cdrage just did that today: https://github.com/ashetty1/kompose/blob/79cc511d1dbe90d3248ff8000f47dbc19e97ecb2/script/test_in_openshift/run.sh#L25

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,229 @@
...
+    # FIXME: Make this generic to cover all cases
+    while [ $(oc get pods | eval ${query_1} | awk '{ print $3 }' | \
+		     grep ${query_1_status} | wc -l) -ne $replica_1 ] ||
+	      [ $(oc get pods | eval ${query_2} | awk '{ print $3 }' | \
+			 grep ${query_2_status} | wc -l) -ne $replica_2 ]; do
+
+	if [ $retry_up -lt 10 ]; then
+	    echo "Waiting for the pods to come up ..."
+	    retry_up=$(($retry_up + 1))
+	    sleep 10
```
cdrage: Can we reduce this to `1` and instead try `120` times (so we wait 2 minutes?)

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,229 @@
...
+function convert::kompose_down_check () {
+    local retry_down=0
+    local pod_count=$1
+    while [ $(oc get pods | wc -l ) != 0 ] &&
+	  [ $(oc get pods | grep -v deploy | grep 'Terminating' | wc -l ) != $pod_count ]; do
+	if [ $retry_down -lt 10 ]; then
+	    echo "Waiting for the pods to go down ..."
+	    oc get pods
+	    retry_down=$(($retry_down + 1))
+	    sleep 10
```
cdrage: Can we reduce this to `1` and instead try `120` times (so we wait 2 minutes?)

cdrage:
```
github.com/kubernetes-incubator/kompose  pr_460 ✔                                                                                                                                                                                                                         3h34m  
▶ make test-openshift 
./script/test_in_openshift/run.sh


===> Starting test <===
Functional tests on OpenShift
./script/test_in_openshift/run.sh: line 29: oc: command not found
FAIL: Please install the oc binary to run testsMakefile:62: recipe for target 'test-openshift' failed
make: *** [test-openshift] Error 1
```

I think there should be a newline after saying the FAIL error.

script/test_in_openshift/lib.sh
ashetty1: Sure

cdrage:
@ashetty1 

Doesn't work it seems if a server isn't already up:

```
github.com/kubernetes-incubator/kompose  pr_460 ✔                                                                                                                                                                                                                        3h36m  ⍉
▶ make test-openshift 
./script/test_in_openshift/run.sh


===> Starting test <===
Functional tests on OpenShift
error: server took too long to respond with version information.
FAIL: Please install the oc binary to run testsMakefile:62: recipe for target 'test-openshift' failed
make: *** [test-openshift] Error 1
```

I believe you should be checking if the binary exists in /usr/local/bin or /usr/bin/ instead of checking via `oc version`

ashetty1:
@cdrage Like this: https://github.com/kubernetes-incubator/kompose/pull/460/commits/79cc511d1dbe90d3248ff8000f47dbc19e97ecb2#diff-627568f7d5722e14bec5cf5a74719fb8L29 ?

cdrage:
@ashetty1 Yes. Checking if OC is there instead of `oc version`.

cdrage:
@ashetty1 

Seems like I'm getting another error:
```
▶ make test-openshift 
./script/test_in_openshift/run.sh


===> Starting test <===
Functional tests on OpenShift
FAIL: oc cluster up failed.
Makefile:62: recipe for target 'test-openshift' failed
make: *** [test-openshift] Error 1
```

There's no output for `oc cluster up` when using `make test-openshift`. May be a good idea in order to diagnose issues.

Of course, there's an error in `oc` (I know what the issue is, was just testing it out):
```
github.com/kubernetes-incubator/kompose  pr_460 ✔                                                                                                                                                                                                                                                                                                                   3h59m  ⍉
▶ oc cluster up
-- Checking OpenShift client ... OK
-- Checking Docker client ... OK
-- Checking Docker version ... FAIL
   Error: Minor number must not contain leading zeroes "03"
```

Which is unrelated, but of course, it'd be nice to have the output from the Makefile so people can diagnose why they can't run `make test-openshift`

cdrage:
Another note is that it would be nice to see the output of `oc cluster up` regardless if it errors or not.

ashetty1:
Cc: @kadel , @cdrage let me know if this looks good now.

cdrage:
Hey @ashetty1 

Tests don't seem to work:
```
github.com/kubernetes-incubator/kompose  pr_460 ✔                                                                                                                                                                                                                                                                                                                      1d  ⍉
▶ make test-openshift 
./script/test_in_openshift/run.sh


===> Starting test <===
Functional tests on OpenShift
-- Checking OpenShift client ... OK
-- Checking Docker client ... OK
-- Checking Docker version ... 
   WARNING: Cannot verify Docker version
-- Checking for existing OpenShift container ... OK
-- Checking for openshift/origin:v3.6.0-alpha.0 image ... OK
-- Checking Docker daemon configuration ... OK
-- Checking for available ports ... OK
-- Checking type of volume mount ... 
   Using nsenter mounter for OpenShift volumes
-- Creating host directories ... OK
-- Finding server IP ... 
   Using 127.0.0.1 as the server IP
-- Starting OpenShift container ... 
   Creating initial OpenShift configuration
   Starting OpenShift using container 'origin'
   Waiting for API server to start listening
   OpenShift server started
-- Adding default OAuthClient redirect URIs ... OK
-- Installing registry ... OK
-- Installing router ... OK
-- Importing image streams ... OK
-- Importing templates ... OK
-- Login to server ... OK
-- Creating initial project "myproject" ... OK
-- Removing temporary directory ... OK
-- Checking container networking ... OK
-- Server Information ... 
   OpenShift server started.
   The server is accessible via web console at:
       https://127.0.0.1:8443

   You are logged in as:
       User:     developer
       Password: developer

   To login as administrator:
       oc login -u system:admin


Testing buildconfig on kompose


Running kompose up ...

INFO Buildconfig using git@github.com:cdrage/kompose.git::pr_460 as source. 
We are going to create OpenShift DeploymentConfigs, Services for your Dockerized application. 

Your application has been deployed to OpenShift. You can run 'oc get dc,svc,is' for details.

Waiting for the pods to come up ...

FAIL: kompose up has failed to bring the pods up
NAME          READY     STATUS              RESTARTS   AGE
foo-1-build   0/1       ContainerCreating   0          2m
```

Most likely to do with the example that you use:
```
version: "2"

services:
    foo:
        build: "./build"
        command: "sleep 3600"
```

Using the "foo" image, and thus the container is never created (stuck on ContainerCreating)

ashetty1:
@cdrage tried running it, and it worked: https://travis-ci.org/ashetty1/kompose/jobs/220470857

I am using the same docker-compose file listed in the examples directory.

cdrage:
@ashetty1 It works for you because you've already built the "foo" image on your machine. 

This fails because you have "foo" listed as the service name and thus OpenShift is trying to find/use an image named "foo" and get's stuck on ContainerCreating because it can't find it.

cdrage:
Have a look at what you linked (https://travis-ci.org/ashetty1/kompose/jobs/220470857) there's even a FAIL in there and it still doesn't fail...

cdrage:
Still get's stuck, here's an extended log:
```
-- Login to server ... OK
-- Creating initial project "myproject" ... OK
-- Removing temporary directory ... OK
-- Checking container networking ... OK
-- Server Information ... 
   OpenShift server started.
   The server is accessible via web console at:
       https://127.0.0.1:8443

   You are logged in as:
       User:     developer
       Password: developer

   To login as administrator:
       oc login -u system:admin


Testing buildconfig on kompose


Running kompose up ...

INFO Buildconfig using git@github.com:cdrage/kompose.git::pr_460 as source. 
We are going to create OpenShift DeploymentConfigs, Services for your Dockerized application. 

Your application has been deployed to OpenShift. You can run 'oc get dc,svc,is' for details.

Waiting for the pods to come up ...

```

```
github.com/kubernetes-incubator/kompose  pr_460 ✔                                                                                                                                                                                                                            4d  
▶ oc get all
bc/foo    Docker    Git@pr_460   1

NAME           TYPE      FROM         STATUS    STARTED   DURATION
builds/foo-1   Docker    Git@pr_460   Pending             

NAME      DOCKER REPO                     TAGS      UPDATED
is/foo    172.30.1.1:5000/myproject/foo             

NAME      REVISION   DESIRED   CURRENT   TRIGGERED BY
dc/foo    0          1         0         config,image(foo:latest)

NAME      CLUSTER-IP   EXTERNAL-IP   PORT(S)     AGE
svc/foo   None         <none>        55555/TCP   1m

NAME             READY     STATUS              RESTARTS   AGE
po/foo-1-build   0/1       ContainerCreating   0          1m
```

@ashetty1 
It also seems that if one test fails, the whole test suite still exits with an exit error 0, which isn't correct.

ashetty1:
@cdrage will reinvestigate. I deleted all the images and ran it again. Still don't see the error:
```
$ oc get all
bc/foo    Docker    Git@master   1

NAME           TYPE      FROM          STATUS     STARTED         DURATION
builds/foo-1   Docker    Git@e8b98e3   Complete   4 minutes ago   4m11s

NAME      DOCKER REPO                         TAGS      UPDATED
is/foo    172.30.184.195:5000/myproject/foo   latest    28 seconds ago

NAME      REVISION   DESIRED   CURRENT   TRIGGERED BY
dc/foo    1          1         1         config,image(foo:latest)

NAME       DESIRED   CURRENT   READY     AGE
rc/foo-1   1         1         1         28s

NAME      CLUSTER-IP   EXTERNAL-IP   PORT(S)     AGE
svc/foo   None         <none>        55555/TCP   4m

NAME             READY     STATUS      RESTARTS   AGE
po/foo-1-build   0/1       Completed   0          4m
po/foo-1-mvjva   1/1       Running     0          26s

```
Will try a few more times just to confirm. The FAIL in the travis log is for `kompose down` which doesn't delete buildconfig pods. The exit status is a problem. Agree, we need to make sure is that make-openshift exits with a non-zero status when there are failures.

cdrage:
This may be a DNS issue on my end with the Docker containers... I'll see if
it is. I remember a similar issue happening a while back with DNS and
retrieving containers.

cdrage:
After all that, found it it was because of this: https://github.com/openshift/origin/issues/13281

Downgrading Docker now and seeing if it fixes the issue.

cdrage:
@ashetty1 

I'm still having troubles trying to replicate this on my local machine.

What version of OC are you running as well as Docker @ashetty1 ?

ashetty1:
@cdrage OC version: v1.4.1
docker on my system is 1.12.1, and on travis is  1.12.3. 

Should I update my docker and try this?

cdrage:
@ashetty1 I'm running 3.6.0 of OpenShift, let me downgrade and see what happens.

cdrage:
Any idea what could be happening @ashetty1 ?

```
NAME          READY     STATUS    RESTARTS   AGE
foo-1-build   0/1       Error     0          12s

github.com/kubernetes-incubator/kompose  pr_460 ✔                                                                                                                                                                                                                            1d  
▶ oc describe po
Name:                   foo-1-build
Namespace:              myproject
Security Policy:        privileged
Node:                   192.168.1.73/192.168.1.73
Start Time:             Wed, 12 Apr 2017 10:58:20 -0400
Labels:                 openshift.io/build.name=foo-1
Status:                 Failed
IP:
Controllers:            <none>
Containers:
  docker-build:
    Container ID:       docker://5d7d419468ac1878c0fe871c360535447a7434767347514f72aac057ac592e6b
    Image:              openshift/origin-docker-builder:v1.4.1
    Image ID:           docker-pullable://openshift/origin-docker-builder@sha256:6a72f5ac83cd9bfa8e61577868232f3405ee3ccd6a51080f12f8aeb5ab38df42
    Port:
    Args:
      --loglevel=0
    State:              Terminated
      Reason:           Error
      Exit Code:        1
      Started:          Wed, 12 Apr 2017 10:58:25 -0400
      Finished:         Wed, 12 Apr 2017 10:58:25 -0400
    Ready:              False
    Restart Count:      0
    Volume Mounts:
      /var/run/docker.sock from docker-socket (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from builder-token-r92lz (ro)
      /var/run/secrets/openshift.io/push from builder-dockercfg-setr8-push (ro)
    Environment Variables:
      BUILD:    {"kind":"Build","apiVersion":"v1","metadata":{"name":"foo-1","namespace":"myproject","selfLink":"/oapi/v1/namespaces/myproject/builds/foo-1","uid":"774895f5-1f90-11e7-b2cb-68f728a040da","resourceVersion":"578","creationTimestamp":"2017-04-12T14:58:19Z","labels":{"buildconfig":"foo","openshift.io/build-config.name":"foo","openshift.io/build.start-policy":"Serial"},"annotations":{"openshift.io/build-config.name":"foo","openshift.io/build.number":"1"}},"spec":{"serviceAccount":"builder","source":{"type":"Git","git":{"uri":"git@github.com:cdrage/kompose.git","ref":"pr_460"},"contextDir":"examples/buildconfig/build"},"strategy":{"type":"Docker","dockerStrategy":{}},"output":{"to":{"kind":"DockerImage","name":"172.30.53.236:5000/myproject/foo:latest"},"pushSecret":{"name":"builder-dockercfg-setr8"}},"resources":{},"postCommit":{},"nodeSelector":null,"triggeredBy":[{"message":"Build configuration change"}]},"status":{"phase":"New","outputDockerImageReference":"172.30.53.236:5000/myproject/foo:latest","config":{"kind":"BuildConfig","namespace":"myproject","name":"foo"}}}

      SOURCE_REPOSITORY:        git@github.com:cdrage/kompose.git
      SOURCE_URI:               git@github.com:cdrage/kompose.git
      SOURCE_CONTEXT_DIR:       examples/buildconfig/build
      SOURCE_REF:               pr_460
      ORIGIN_VERSION:           v1.4.1+3f9807a
      PUSH_DOCKERCFG_PATH:      /var/run/secrets/openshift.io/push
Conditions:
  Type          Status
  Initialized   True 
  Ready         False 
  PodScheduled  True 
Volumes:
  docker-socket:
    Type:       HostPath (bare host directory volume)
    Path:       /var/run/docker.sock
  builder-dockercfg-setr8-push:
    Type:       Secret (a volume populated by a Secret)
    SecretName: builder-dockercfg-setr8
  builder-token-r92lz:
    Type:       Secret (a volume populated by a Secret)
    SecretName: builder-token-r92lz
QoS Class:      BestEffort
Tolerations:    <none>
Events:
  FirstSeen     LastSeen        Count   From                    SubobjectPath                   Type            Reason          Message
  ---------     --------        -----   ----                    -------------                   --------        ------          -------
  16s           16s             1       {default-scheduler }                                    Normal          Scheduled       Successfully assigned foo-1-build to 192.168.1.73
  12s           12s             1       {kubelet 192.168.1.73}  spec.containers{docker-build}   Normal          Pulling         pulling image "openshift/origin-docker-builder:v1.4.1"
  11s           11s             1       {kubelet 192.168.1.73}  spec.containers{docker-build}   Normal          Pulled          Successfully pulled image "openshift/origin-docker-builder:v1.4.1"
  10s           10s             1       {kubelet 192.168.1.73}  spec.containers{docker-build}   Normal          Created         Created container with docker id 5d7d419468ac; Security:[seccomp=unconfined]
  10s           10s             1       {kubelet 192.168.1.73}  spec.containers{docker-build}   Normal          Started         Started container with docker id 5d7d419468ac
```

ashetty1:
@cdrage what do the logs say? Does it say something like this?
```
# oc get pods
NAME          READY     STATUS    RESTARTS   AGE
foo-1-build   0/1       Error     0          1m

# oc logs foo-1-build
Cloning "https://github.com/kubernetes-incubator/kompose.git" ...
error: build error: error: pathspec 'pr_460' did not match any file(s) known to git.

```

ashetty1:
@surajssd, have sent in the empytvols changes.

ashetty1:
@cdrage I investigated a bit and realised that build config usually errors out because due to git issues. 

I tried on a fresh fedora 25 machine, and ran the tests; it seems to work:

```
# oc get all
bc/foo    Docker    Git@master   1

NAME           TYPE      FROM          STATUS     STARTED         DURATION
builds/foo-1   Docker    Git@0464d24   Complete   4 minutes ago   3m47s

NAME      DOCKER REPO                     TAGS      UPDATED
is/foo    172.30.1.1:5000/myproject/foo   latest    About a minute ago

NAME      REVISION   DESIRED   CURRENT   TRIGGERED BY
dc/foo    1          1         1         config,image(foo:latest)

NAME       DESIRED   CURRENT   READY     AGE
rc/foo-1   1         1         1         1m

NAME      CLUSTER-IP   EXTERNAL-IP   PORT(S)     AGE
svc/foo   None         <none>        55555/TCP   4m

NAME             READY     STATUS      RESTARTS   AGE
po/foo-1-build   0/1       Completed   0          4m
po/foo-1-tzkk7   1/1       Running     0          53s


# docker version
Client:
 Version:         1.12.6
 API version:     1.24
 Package version: docker-common-1.12.6-6.gitae7d637.fc25.x86_64
 Go version:      go1.7.4
 Git commit:      ae7d637/1.12.6
 Built:           Mon Jan 30 16:15:28 2017
 OS/Arch:         linux/amd64

Server:
 Version:         1.12.6
 API version:     1.24
 Package version: docker-common-1.12.6-6.gitae7d637.fc25.x86_64
 Go version:      go1.7.4
 Git commit:      ae7d637/1.12.6
 Built:           Mon Jan 30 16:15:28 2017
 OS/Arch:         linux/amd64
[root@hp-z600-02 buildconfig]# oc versiob
Error: unknown command "versiob" for "oc"

# oc version
oc v1.5.0-rc.0+49a4a7a
openshift v1.5.0-rc.0+49a4a7a
kubernetes v1.5.2+43a9be4

```

cdrage:
@ashetty1 

Still not working, this is the error I get from foo-1-build:

```
github.com/kubernetes-incubator/kompose  pr_460 ✔                                                                                                                                                                                                                         2h57m  
▶ oc logs foo-1-build
error: failed to retrieve cgroup limits: cannot determine cgroup limits: open /sys/fs/cgroup/cpu/cpu.cfs_quota_us: no such file or directory
```

At this point, it's possible it's an issue with Debian + Docker + Kubernetes... Judging from this issue: https://github.com/openshift/origin/issues/8074


@kadel @surajssd @containscafeine can you guys test this PR and give your LGTM?

At this point, this looks good to me.

cdrage:
(other than of course squashing the commits)

surajssd:
blocked by https://github.com/kubernetes-incubator/kompose/issues/382

surajssd:
It works for me except for the kompose down issue in #382 
otherwise it's good

procrypt:
@ashetty1 I was looking into the issue why `pods` are not getting deleted, I found out that the label `io.kompose.service` is not getting populated into the `podSpec` when we use `buildConfig` and since we are deleting the objects based on labels, `kompose down` fails to delete the `pods`.

`podSpec` of pods created with `buildConfig`
```bash
$ oc describe pods foo-1-build 
Name:                   foo-1-build
Namespace:              ns
Security Policy:        privileged
Node:                   192.168.121.71/192.168.121.71
Start Time:             Thu, 20 Apr 2017 11:30:02 +0000
Labels:                 openshift.io/build.name=foo-1
Status:                 Failed
IP:                     172.17.0.2
Controllers:            <none>
```

As you can see there is different label `openshift.io/build.name=foo-1` not `io.kompose.service`, this is the reason why `kompose down` fails to delete it.

The issue is with `OpenShift` so I don't how long it will take to resolve this.

ashetty1:
@cdrage , @surajssd if the issue is with OpenShift, should we unblock and merge this PR?

surajssd:
@ashetty1 I would recommend commenting that test case, create an issue that tracks that commented test case and then I guess we can proceed.

Once we know how to delete that undeleteed pod we uncomment that test.

cdrage:
@ashetty1 Got to make sure that even though pods don't get deleted, that after `oc cluster down`.

ashetty1:
@surajssd have taken buildconfig tests off this commit.

surajssd:
@ashetty1 please correct me if I am wrong here, but as I see it there are two issues with this PR.

**First issue**: there is a problem with `buildconfig` that even if `buildconfig` has label the build pod does not have the `label`. So it is not deleted. And it is tracked at https://github.com/kubernetes-incubator/kompose/issues/382

So I would recommend that we keep this test, and just comment it out. In issue https://github.com/kubernetes-incubator/kompose/issues/572 we say that we need to uncomment this test once issue https://github.com/kubernetes-incubator/kompose/issues/382 is fixed.

**Second issue**: another problem with buildconfig is that if the git repo is on some random commit the buildconfig generated is weird and it fails. And it is tracked at https://github.com/kubernetes-incubator/kompose/issues/561

Here what we can do is not rely on the test case example in the repo but just git clone some git repo in temporary directory and then try to run buildconfig test there so that we will get consistent results. So this does not need any special issue to track anything.

ashetty1:
@surajssd I have just commented out the `kompose down` part now for buildconfig tests.

script/test_in_openshift/tests/buildconfig.sh
```diff
@@ -0,0 +1,35 @@
...
+docker_compose_file="${KOMPOSE_ROOT}/examples/buildconfig/docker-compose.yml"
+
+# Run kompose up
+convert::kompose_up $docker_compose_file
+
+# Check if the pods are up.
+convert::kompose_up_check -p foo
+
+# Kompose down for buildconfig fails being tracked at #382
+# convert::kompose_down $docker_compose_file
```
surajssd: this is the one you have commented right? @ashetty1

surajssd:
@cdrage this looks good to me see if you are good with it go ahead and merge it

cdrage:
Good enough! Although a new issue will have to be opened to re-add buildconfig testing once the other PR has been merged.

surajssd:
@cdrage so there is one issue for this, that needs some work, i was gonna point to the line number of the merged code sayig uncomment this or something like that once this is merged in here https://github.com/kubernetes-incubator/kompose/issues/572

script/test_in_openshift/tests/buildconfig.sh
ashetty1: Yes, that's right.

ashetty1:
@kadel, could you please take a look?

script/test_in_openshift/lib.sh
kadel: yep, `install_oc_client` should be run only if its run within travis. If you make sure that its run only there than its ok.

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,231 @@
...
+
+    if [ -z $replica ]; then
+       replica_1=1
+       replica_2=1
+    else
+       replica_1=$replica
+       replica_2=$replica
+    fi
+    
+    pod_1=$( echo $pod | awk '{ print $1 }')
```
kadel: this means that you can't check more than two pods?

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,231 @@
...
+}
+
+function install_oc_client () {
+    # Valid only for Travis
+    convert::print_msg "Installing oc client binary ..."
+    sudo sed -i 's:DOCKER_OPTS=":DOCKER_OPTS="--insecure-registry 172.30.0.0/16 :g' /etc/default/docker
+    sudo mv /bin/findmnt /bin/findmnt.backup
+    sudo /etc/init.d/docker restart
+    # FIXME
+    wget https://github.com/openshift/origin/releases/download/v1.4.1/openshift-origin-client-tools-v1.4.1-3f9807a-linux-64bit.tar.gz -O /tmp/oc.tar.gz 2> /dev/null > /dev/null
```
kadel: v1.5 was released, lets use that

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,231 @@
...
+    if [ $(oc get pods | eval ${query_1} | awk '{ print $3 }' | \
+		  grep ${query_1_status} | wc -l) -eq $replica_1 ] &&
+	   [ $(oc get pods | eval ${query_2} | awk '{ print $3 }' | \
+		      grep ${query_2_status} | wc -l) -eq $replica_2 ]; then
+	oc get pods
+	convert::print_pass "All pods are Running now. kompose up is successful."
+    fi
+}
+
+function convert::kompose_down_check () {
```
kadel: can you please add comment explaining what parameter is required for this function and what it means?

script/test_in_openshift/lib.sh
ashetty1: For now, yes.

kadel: I don't like that :-(
If you are not going to fix that, please write some big warning that this is the limitation and also TODO that this needs to be fixed. It will cause a lot of confusion for anyone writing tests.

script/test_in_openshift/lib.sh
```diff
@@ -0,0 +1,235 @@
...
+    exit_status=$?
+
+    if [ $exit_status -ne 0 ]; then
+	FAIL_MSGS=$FAIL_MSGS"kompose down failed with exit status: $exit_status\n";
+	exit $exit_status;
+    fi
+}
+
+
+function convert::kompose_up_check () {
```
kadel: This function can cause a lot of confusions, can you please add comment explaining what are the limitations of this function and what it does?

kadel:
One last stupid question :innocent: 
Why is every function prefixed with `convert::`?

ashetty1:
@kadel just used the convention that existed. No other reason :)

Have made the changes you requested.

kadel:
> @kadel just used the convention that existed. No other reason :)

OK, it looks a bit strange there, but apparently  no one else had has any issues with that, so lets keep it ;-)

Thank you for changes :+1: 

LGTM

PR #465, opened 2017-03-01
Update release script

cdrage:
Had an issue where I don't git pull / merge with master after the PR has
been merged.

This adds an option to sync with master before continuing.

### DIFF
--- a/script/release.sh
+++ b/script/release.sh
@@ -71,6 +71,7 @@ clone() {
   cd $CLI
   git remote remove origin
   git remote add origin git@github.com:$ORIGIN_REPO/$CLI.git
+  git remote add upstream http://github.com/$UPSTREAM_REPO/$CLI
   git checkout -b release-$1
   cd ..
 }
@@ -132,6 +133,13 @@ git_pull() {
   cd $CLI
   git pull
   cd ..
+}
+
+
+git_sync() {
+  cd $CLI
+  git fetch upstream master
+  git rebase upstream/master
 }
 
 push() {
@@ -217,6 +225,7 @@ main() {
   "Generate changelog"
   "Generate GitHub changelog"
   "Create PR"
+  "Sync with master"
   "Build binaries"
   "Upload the binaries and push to GitHub release page"
   "Clean"
@@ -239,6 +248,9 @@ main() {
               ;;
           "Create PR")
               git_commit $VERSION
+              ;;
+          "Sync with master")
+              git_sync
               ;;
           "Build binaries")
               build_binaries


script/release.sh
```diff
@@ -134,6 +134,13 @@ git_pull() {
   cd ..
 }
 
+
+git_sync() {
+  cd $CLI
+  git pull origin
```
kadel: I'm not sure about this :-( 
Isn't origin pointing to fork? It should pull from upstream no?

cdrage: Ah yeah, you're right. Should be pulling from upstream.

cdrage:
@kadel Updated :)

script/release.sh
```diff
@@ -71,6 +71,7 @@ clone() {
   cd $CLI
   git remote remove origin
   git remote add origin git@github.com:$ORIGIN_REPO/$CLI.git
+  git remote add upstream git@github.com:$UPSTREAM_REPO/$CLI.git
```
kadel: Can we maybe https instead of ssh? It will prevent accidental pushes to upstream.

script/release.sh
```diff
@@ -134,6 +135,13 @@ git_pull() {
   cd ..
 }
 
+
+git_sync() {
+  cd $CLI
+  git pull upstream
```
kadel: I still think this won't work. You should also specify branch.
`git pull upstream master` I would probably use fetch here. `git fetch upstream master`

script/release.sh
cdrage: I'd prefer it SSH to be honest... Having to enter http credentials each-time is a kill-joy. I have no arguments to say that ssh is any better haha. Just personal preference of mine.

script/release.sh
cdrage: Hmm. It *should* work, I just tested it out and it pulled an updated branch of gh-pages I recently commited to. I'll update the script!

script/release.sh
kadel: But you never push to upstream.

This is why I suggested https, so auth prompt pops-up and you know that you are pushing to wrong remote.

script/release.sh
kadel: I've tried  and got:
```
You asked to pull from the remote 'upstream', but did not specify
a branch. Because this is not the default configured remote
for your current branch, you must specify a branch on the command line.
```

script/release.sh
cdrage: Ahhhhh, I get it. So https for upstream and ssh for origin and then we have the best of both-worlds.

PR #468, opened 2017-03-02
Add cap_add and cap_drop to unsupported keys

kadel:
Kompose reads `cap_add` and `cap_drop` from docker-compose.yml but than it is not used anywhere.


cdrage:
LGTM :+1:

PR #492, opened 2017-03-14
Add install instruction for MacOS X (using brew)

kadel:
Kompose is homebrew - https://github.com/Homebrew/homebrew-core/blob/master/Formula/kompose.rb


cdrage:
Can't confirm it works (don't have a Mac!) But LGTM!

PR #497, opened 2017-03-17
travis-ci: send report to coveralls in after_success section

kadel:
Connection to coveralls.io sometimes fails and its marking tests as fail
even then are OK. Exit code from after_success do not affects build
result.


cdrage:
LGTM

PR #498, opened 2017-03-17
do not fail if there is a golint violation

kadel:
golint output is just suggestion. Show output, but don't fail.
This also remove unnecessary bash script that is no longer needed.


cdrage:
LGTM

PR #499, opened 2017-03-20
Added dockerfile key support

### LINKED ISSUE #486: kompose should respect the dockerfile key in docker-compose.yml
If one doesn't use the default Dockerfile, but specified it using dockerfile: in the yml, the generated build config does not have a dockerFilePath:  under dockerStrategy.  This is in version 0.3.0 (HEAD)

surajnarwade:
Fixes #486
This commit will add `dockerfilepath` key under Dockerstratergy in Buildconfig.
dockerfilepath allow us to use dockerfiles which are named different than `Dockerfile`

### DIFF
--- a/pkg/kobject/kobject.go
+++ b/pkg/kobject/kobject.go
@@ -84,6 +84,7 @@ type ServiceConfig struct {
 	Tty           bool                `compose:"tty" bundle:""`
 	MemLimit      yaml.MemStringorInt `compose:"mem_limit" bundle:""`
 	TmpFs         []string            `compose:"tmpfs" bundle:""`
+	Dockerfile    string              `compose:"dockerfile" bundle:""`
 }
 
 // EnvVar holds the environment variable struct of a container
--- a/pkg/loader/compose/compose.go
+++ b/pkg/loader/compose/compose.go
@@ -76,7 +76,6 @@ func checkUnsupportedKey(composeProject *project.Project) []string {
 		"Uts":           false,
 		"ReadOnly":      false,
 		"Ulimits":       false,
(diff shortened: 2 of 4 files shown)


surajnarwade:
cc @kadel @surajssd @cdrage

cdrage:
@surajnarwade 

Wouldn't it have to be removed from: https://github.com/kubernetes-incubator/kompose/blob/master/pkg/loader/compose/compose.go#L80 as well?

cdrage:
Tests need to be added too (cmd tests)

kadel:
> Wouldn't it have to be removed from: https://github.com/kubernetes-incubator/kompose/blob/master/pkg/loader/compose/compose.go#L80 as well?

yes, It shouldn't be there in the first place :-D This only tracks top level service keys, dockerfile is hidden in build key

surajnarwade:
@cdrage , for cmd-test, I have add context, which will fail the test as PR https://github.com/kubernetes-incubator/kompose/pull/454 is blocked, once that PR is merged, I will add functional test for this one.

cdrage:
Few spelling errors in the commit message:
```
Fixes #486
This commit will add `dockerfilepath` key under Dockerstratergy in Buildconfig.
dockerfilepath allow us to use dockerfiles which are named different than `Dockerfile`
```

Please give an example as well in the commit message / elaborate :)

kadel:
needs rebase, but otherwise lgtm

surajnarwade:
@kadel , done with rebase

cdrage:
@surajnarwade thank you for updating the commit message. LGTM! :+1:

PR #507, opened 2017-03-23
Update the readme with tarball and installation instructions

cdrage:
Updates the readme to reflect that we now have a tarball as well as
latest development instructions on prefering the binary installation
over `go get`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,13 +33,7 @@ Other examples are provided in the _examples_ [directory](./examples).
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


kadel:
It looks like something went wrong with you branch. 0f7c13b00f570239f6afa6ec9c3858f2595c999c shouldn't be in this pr

cdrage:
Crap, my mistake. My "master" branch was two commits ahead since I was testing against #500 

Should be corrected / updated now.

PR #529, opened 2017-03-31
Update vendoring + fix issue with timestamps being added to log

cdrage:
As per the vendoring here:
https://github.com/kubernetes-incubator/kompose/commit/e9544ca89477ed76bd75f1160f9a89686f75e95d
for some odd issue, text_formatter.go in the logrus package was updated
to an older file.

This commit updates the vendoring as well as specifies a specific
version for logrus to be used.


cdrage:
Closes #529

sinnykumari:
Thanks Charlie! This commit looks good to me in terms of /vendor/golang.org/x/sys update and should fix issue#532

dustymabe:
@cdrage, do you have any comprehensive tests you can run?

dustymabe:
@cdrage can you add that this fixes #532 to the commit comment?

also I have a question, if we are updating a lot of different things then I think we might need a new release of kompose in order to put it out. would rather not pull back a patch if its going to be a large change.

cdrage:
@dustymabe No comprehensive tests are being run right now, we're working on adding more integration tests. But unit tests + cmd tests are the only ones being ran at the moment.

Yup. We can do another release on Monday!

dustymabe:
sounds good. once you commit PR we can have @surajssd build an rpm from this commit (don't need a release yet) and I can run it through the fedora build system to verify everything works on ppc **before you do a release**.

cdrage:
I've confirmed that this commit works and we've gotten a LGTM.

Let's merge this!

PR #528, opened 2017-03-31
Fix the examples

cdrage:
Moves `docker-compose.yaml` to `docker-compose-counter.yaml` and
replaces the default docker-compose.yaml with the example from the
Kompose.io website.

We also rename each file from yml to yaml.

As well as do some slight modifications / formatting to
docker-compose-bundle.dab

PR #531, opened 2017-03-31
Move docker-compose.yml test file to tests

cdrage:
Moves a file that was incorrectly being used from examples to
script/test/fixtures.


cdrage:
Unblocks #528

kadel:
it might be better to move ` script/test/docker-compose.yml` to its own directory in `fixtures` so its inline with other tests

cdrage:
@kadel Done! Moved it to `redis-example`.

cdrage:
There, tests should pass now!

cdrage:
Woo :+1:

PR #533, opened 2017-03-31
Only ignore the docker-compose yaml file in the root directory

cdrage:
Woops! It was ignoring all of them in the directory. We should only be
avoiding the root dir docker-compose files!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,8 @@
 # Ignore compiled Kompose files
 kompose
 bin
-docker-compose.yaml
-docker-compose.yml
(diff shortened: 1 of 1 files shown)


surajnarwade:
LGTM :+1:

kadel:
` examples/docker-compose.yaml` and `script/test/fixtures/docker-compose.yml` were deleted before? What are those? Maybe leftovers from https://github.com/kubernetes-incubator/kompose/pull/528 ?

cdrage:
@kadel Yup, they were, I rebased and removed them.

PR #535, opened 2017-04-03
Revert "Kompose will keep trying its job"

cdrage:
Reverts kubernetes-incubator/kompose#477


cdrage:
ping @kadel 

Let's revert this for now since we're releasing Kompose today. Don't want this in / extra-test the PR.

PR #536, opened 2017-04-03
new take on "Kompose will keep trying its job #477"

kadel:
This is reopening of #477 there were some issues that we forgot to address. 

Issue was that if it can't connect to cluster it panics.
Problem is that we continue after every error.
There should be return after some errors, and after every error in switch should be break so we don't 
try to get reaper using client which creation field or calling `Stop` on reaper that field to create.

I've added new commit on top of  original @surajnarwade's commit, so its easier to review.
Once we agreed that this is the way we wan't to do it we can squash it

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -695,14 +695,15 @@ func (k *Kubernetes) Undeploy(komposeObject kobject.KomposeObject, opt kobject.C
 	var errorList []error
 	//Convert komposeObject
 	objects, err := k.Transform(komposeObject, opt)
-
 	if err != nil {
 		errorList = append(errorList, err)
+		return errorList
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -691,17 +691,19 @@ func (k *Kubernetes) Deploy(komposeObject kobject.KomposeObject, opt kobject.Con
...
-		return errors.Wrap(err, "k.Transform failed")
+		errorList = append(errorList, err)
+		return errorList
 	}
 
 	client, namespace, err := k.GetKubernetesClient()
 	if err != nil {
-		return err
+		errorList = append(errorList, err)
+		return errorList
```
cdrage: This is better! :+1:

cdrage:
LGTM

PR #551, opened 2017-04-07
Update document describing release process

kadel:
Updated release process.
@cdrage, does is describe how you do it?
Is there something important that I forgot?

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,9 +1,8 @@
 # Release Process
 
-The Kubernetes Template Project is released on an as-needed basis. The process is as follows:
+The process is as follows:
 
-1. An issue is proposing a new release with a changelog since the last release
-1. All [OWNERS](OWNERS) must LGTM this release
+1. A PR proposing a new release with a changelog since the last release
+1. At least 2 or more [OWNERS](OWNERS) must LGTM this release
+1. The release PR is closed
 1. An OWNER runs `git tag -s $VERSION` and inserts the changelog and pushes the tag with `git push $VERSION`
-1. The release issue is closed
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -1,9 +1,8 @@
...
 
-The Kubernetes Template Project is released on an as-needed basis. The process is as follows:
+The process is as follows:
 
-1. An issue is proposing a new release with a changelog since the last release
-1. All [OWNERS](OWNERS) must LGTM this release
-1. An OWNER runs `git tag -s $VERSION` and inserts the changelog and pushes the tag with `git push $VERSION`
-1. The release issue is closed
-1. An announcement email is sent to `kubernetes-dev@googlegroups.com` with the subject `[ANNOUNCE] kubernetes-template-project $VERSION is released`
+1. An PR is proposing a new release with a changelog since the last release
```
cdrage: `A PR proposing a new release`

kadel: > I believe we should put down the processing of running the release.sh script, no?

yes

> I can elaborate on this and make this document more extensive (go into a lot of detail in regards to how we do the release and how to replicate it)

That would be great. I just wanted to update it to something that is at least partial true ;-) 

If what I wrote isn't  complete nonsense we could merge it, it will be better than what was there before, than you can expand/rewrite it later. What do you think?

kadel:
> I believe we should put down the processing of running the release.sh script, no?

yes

> I can elaborate on this and make this document more extensive (go into a lot of detail in regards to how we do the release and how to replicate it)

That would be great. I just wanted to update it to something that is at least partial true ;-) 

If what I wrote isn't  complete nonsense we could merge it, it will be better than what was there before, than you can expand/rewrite it later. What do you think?

cdrage:
@kadel 
I agree, let's move fast and merge this for now until we go with some more doc updates!

PR #562, opened 2017-04-20
Updating Vendor

surajnarwade:
As https://github.com/docker/libcompose/pull/457 and
https://github.com/docker/libcompose/pull/450 are merged now in libcompose


surajnarwade:
it's failing because, as this PR and https://github.com/kubernetes-incubator/kompose/pull/454 depend on each other, any one them needs to be merged before.

PR #568, opened 2017-04-24
fix output replication-controller and daemon-set

nkysg:
@cdrage  fix output   replication-controller and daemon-set


cdrage:
@nkysg Thanks! :D LGTM :dancing_women:

Issue #538, opened 2017-04-03
add documentation about renaming service


kadel:
We should add warning to docs explaining why, when and how we rename service as it may break some docker-compose files

see https://github.com/kubernetes-incubator/kompose/pull/509

surajnarwade:
I will pick this :)

procrypt:
@surajnarwade I'm working on this

procrypt:
Sorry I didn't mentioned it here.

surajnarwade:
@procrypt no issue :)

PR #569, opened 2017-04-24
Add warning about change in the service name.

procrypt:
Add warning about change in the service name of docker-compose file that contains `_` in it to `-`.

Fixes: #538 

cc: @cdrage @surajssd @surajnarwade @containscafeine

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -394,5 +394,10 @@ services:
 ```
 
 
-#### Warning about Deployment Config
+#### Warning about Deployment Config's
+
 If the Docker Compose file has a volume specified for a service, the Deployment (Kubernetes) or DeploymentConfig (OpenShift) strategy is changed to "Recreate" instead of "RollingUpdate" (default). This is done to avoid multiple instances of a service from accessing a volume at the same time.
+
+If the Docker Compose file has service name with `_` in it (eg.`web_service`), then it will be replaced by `-` and the service name will be renamed accordingly (eg.`web-service`). Kompose does this because "Kubernetes" doesn't allow `_` in object name.
+
+Please note that changing service name might break some `docker-compose` files.


docs/user-guide.md
```diff
@@ -396,3 +396,10 @@ services:
 
 #### Warning about Deployment Config
 If the Docker Compose file has a volume specified for a service, the Deployment (Kubernetes) or DeploymentConfig (OpenShift) strategy is changed to "Recreate" instead of "RollingUpdate" (default). This is done to avoid multiple instances of a service from accessing a volume at the same time.
+
+#### Warning about change of Service name
+If the Docker Compose file has service name with `_` in it (eg.`web_service`), then it will be replaced by `-` and the service name will be renamed accordingly (eg.`web-service`).
+
+**We are doing this because "Kubernetes" doesn't allow `_` in object name.**
```
pradeepto: This looks like another heading when it is rendered. Can you use different markup instead? 

Also, instead of "we are doing this", do you think "kompose does this" will be more appropriate?

procrypt: @pradeepto Thanks for review, I'll update the doc.

procrypt:
@pradeepto PTAL.

pradeepto:
LGTM but wait for someone else to +1. Then feel free to merge. cc @surajnarwade  @surajssd @containscafeine @kadel @cdrage

docs/user-guide.md
```diff
@@ -396,3 +396,10 @@ services:
 
 #### Warning about Deployment Config
 If the Docker Compose file has a volume specified for a service, the Deployment (Kubernetes) or DeploymentConfig (OpenShift) strategy is changed to "Recreate" instead of "RollingUpdate" (default). This is done to avoid multiple instances of a service from accessing a volume at the same time.
+
+#### Warning about change of Service name
+If the Docker Compose file has service name with `_` in it (eg.`web_service`), then it will be replaced by `-` and the service name will be renamed accordingly (eg.`web-service`).
+
+##### Kompose does this because "Kubernetes" doesn't allow `_` in object name.
```
cdrage: I think this shouldn't have a headline. Just remove the `####` and add another space so it's separate / paragraphed.

procrypt: @cdrage Ack.

surajnarwade:
LGTM :+1:

docs/user-guide.md
```diff
@@ -396,3 +396,11 @@ services:
...
 #### Warning about Deployment Config
 If the Docker Compose file has a volume specified for a service, the Deployment (Kubernetes) or DeploymentConfig (OpenShift) strategy is changed to "Recreate" instead of "RollingUpdate" (default). This is done to avoid multiple instances of a service from accessing a volume at the same time.
+
+Warning about change of Service name.
+
+If the Docker Compose file has service name with `_` in it (eg.`web_service`), then it will be replaced by `-` and the service name will be renamed accordingly (eg.`web-service`).
+
+##### Kompose does this because "Kubernetes" doesn't allow `_` in object name.
+
+Please note that changing service name might break some `docker-compose` files.
```
cdrage: I think there was some confusion, all of this should be:
```
#### Warning about Deployment Config's

If the Docker Compose file has a volume specified for a service, the Deployment (Kubernetes) or DeploymentConfig (OpenShift) strategy is changed to "Recreate" instead of "RollingUpdate" (default). This is done to avoid multiple instances of a service from accessing a volume at the same time.

If the Docker Compose file has service name with `_` in it (eg.`web_service`), then it will be replaced by `-` and the service name will be renamed accordingly (eg.`web-service`). Kompose does this because "Kubernetes" doesn't allow `_` in object name.

Please note that changing service name might break some `docker-compose` files.
```

cdrage: ping @procrypt

procrypt: @cdrage Updated.

cdrage:
LGTM

PR #581, opened 2017-04-28
Update the release script again

cdrage:
Updates the release script to use the actual GOPATH directory as
previous problems with commit hashes / versions in `kompose version`.

Cleans this up in regards to all the `cd` and `cd ..` commands.

No need to `git clone`.


cdrage:
Going to go ahead and merge this before the release today (so we can properly run this script).

PR #612, opened 2017-05-19
Update `kompose completion`

cdrage:
Updates Kompose completion with neutral language (using shell instead of
bash).

Fixes the indentation issues when outputting `kompose completion --help`


surajnarwade:
lgtm :+1:

PR #613, opened 2017-05-23
Fix incorrect tag in BuildConfig.

### LINKED ISSUE #611: image tag name absent from generated imagestream config
On running `kompose up --provider openshift` the image build is completed, but the container never starts. All I see on `oc get pods` is:

```
myproject   web-1-build                     0/1       Completed           0          10m
```

On manually triggering the container through `oc rollout latest dc/web`, I get 
```
Error from server (BadRequest): cannot trigger a deployment for "web" because it contains unresolved images
```
On checking `oc edit is web` I get 
```
spec: {}
```

so then on checking `oc get is` I see:
```
web           172.30.1.1:5000/myproject/web           latest    20 minutes ago
```
But my docker compose file contains the following info:

```
<snip>
  web:
    restart: always
    build:
      context: lib/backend
      args:
        BACKEND_SERVER_PORT: $BACKEND_SERVER_PORT
    env_file: .env
    image: sarjitsu:backend
    ports:
    - "${BACKEND_PORT_MAPPING}:${BACKEND_SERVER_PORT}"
    depends_on:
...
</snip>
```

The `:backend` tag should've been applied to the image, but that field `spec` in `is` is empty. Hence we see `..because it contains unresolved images` and no container launches.

We could workaround this through `oc edit dc web` ..by changing `name: web:backend` to `name: web:latest`  like this:

```
  test: false
  triggers:
  - type: ConfigChange
  - imageChangeParams:
      automatic: true
      containerNames:
      - web
      from:
        kind: ImageStreamTag
        name: web:latest
        namespace: myproject
      lastTriggeredImage: 172.30.1.1:5000/myproject/web@sha256:a906e1715d23ac84f431c565cd32ee148b997d49bf8b5a529351f2a1c2a812e7
    type: ImageChange
```

Conclusion is, the tag `:backend` is not appearing in image stream, but is present in deployment config.

The same error could be verified by checking contents of `web-imagestream.yaml`  when generating config files `kompose convert -f docker-compose.yml  --provider openshift` ..where the `:backend` tag is absent :
```
apiVersion: v1
kind: ImageStream
metadata:
  creationTimestamp: null
  labels:
    io.kompose.service: web
  name: web
spec: {}
status:
  dockerImageRepository: ""
``` 

#### kompose
0.6.0 (HEAD)
(built from repo's master branch)

#### openshift
```
oc v1.5.0
kubernetes v1.5.2+43a9be4
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://10.209.68.209:8443
openshift v1.5.0+031cbe4
kubernetes v1.5.2+43a9be4
```

### Supplementary info
Docker compose config being referred to in here, is at https://github.com/arcolife/sarjitsu/blob/non-root/docker-compose.yml#L56

### Related

Not sure this should be the way it is, but a tangentially related issue is that [this image name `sarjitsu:backend`](https://github.com/arcolife/sarjitsu/blob/non-root/docker-compose.yml#L63) is not appearing anywhere in openshift's imagestreams / pod names, while It appears that way when using docker-compose:
```
sarjitsu_web_1           /docker-entrypoint.sh backend    Up      0.0.0.0:9704->80/tcp   
```

Rather, it's all based out of the service names I assign, like `web:backend`

cc/ @procrypt @surajnarwade @containscafeine

kadel:
For services with build and image keys DeploymentConfig respects tag from docker-compose image key.
But BuildConfig image tag was always set to  'latest'.
Result of this was that deployment wasn't trigired after successful build.
This fixes it by setting BuildConfig output image tag to the same tag that is used for  DeploymentConfig (tag from docker-compose image key)

fixes #611

### DIFF
--- a/pkg/transformer/openshift/openshift.go
+++ b/pkg/transformer/openshift/openshift.go
@@ -245,7 +245,7 @@ func initBuildConfig(name string, service kobject.ServiceConfig, repo string, br
 				Output: buildapi.BuildOutput{
 					To: &kapi.ObjectReference{
 						Kind: "ImageStreamTag",
-						Name: name + ":latest",
+						Name: name + ":" + getImageTag(service.Image),
 					},
 				},
 			},
--- a/pkg/transformer/openshift/openshift_test.go
+++ b/pkg/transformer/openshift/openshift_test.go
@@ -281,43 +281,67 @@ func TestGetAbsBuildContext(t *testing.T) {
 
 // Test initializing buildconfig for a service
 func TestInitBuildConfig(t *testing.T) {
-	dir := testutils.CreateLocalGitDirectory(t)
-	testutils.CreateSubdir(t, dir, "a/build")
-	defer os.RemoveAll(dir)
-
 	serviceName := "serviceA"
-	repo := "https://git.test.com/org/repo"
+	repo := "https://git.test.com/org/repo1"
 	branch := "somebranch"
 	buildArgs := []kapi.EnvVar{{Name: "name", Value: "value"}}
 	value := "value"
-	sc := kobject.ServiceConfig{
-		Build:      filepath.Join(dir, "a/build"),
-		Dockerfile: "Dockerfile-alternate",
-		BuildArgs:  map[string]*string{"name": &value},
-	}
-	bc, err := initBuildConfig(serviceName, sc, repo, branch)
-	if err != nil {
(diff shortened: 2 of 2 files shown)


kadel:
This PR includes small refactoring for `TestInitBuildConfig` function.
The way it was done before was making it hard to add new test for new services.

cdrage:
LGTM :+1:

surajnarwade:
LGTM :+1:

PR #618, opened 2017-05-25
Update version number in setup.md file in release script

cdrage:
This change updates setup.md to include the new version number when
doing a release.


cdrage:
Mergin' before today's release! Just a simple update to release.sh

PR #623, opened 2017-05-26
Move logo to the bottom

cdrage:
Moves the logo to the bottom of the sidebar rather than the top.


cdrage:
Looks better IMO. Doesn't indent the title / not have a full-logo.

surajnarwade:
:+1: , Nice looking now :)

PR #629, opened 2017-06-02
Fix adding Jekyll format to the top

cdrage:
Problem with `/permalink/`. This adds a few bash lines that updates
index.md without the permalink.


cdrage:
Unfortunately I'm going to have to merge this in, in order to troubleshoot what's up with GitHub. If this doesn't work i'll revert it manually.

PR #634, opened 2017-06-08
Update widgets (add godoc) as well as README

cdrage:
Update the widgets (as well as add godoc), and update the README, making
the `kompose completion` section more concise.


cdrage:
ping @kadel

Issue #652, opened 2017-06-19
Deprecation warnings are not showing


cdrage:
See: https://github.com/spf13/cobra/issues/463

Marking flags as `MarkDeprecated` does not show the proper warning.

Issue #390, opened 2017-01-24
Docker DAB support does not currently work (still in experimental)


cdrage:
Converting the current example we have produces this error:
```
▶ ./kompose convert -f examples/docker-compose-bundle.dab
ERRO Could not parse config for project examples : yaml: unmarshal errors:
  line 24: cannot unmarshal !!str `0.1` into config.RawService 
FATA Failed to load compose file: yaml: unmarshal errors:
  line 24: cannot unmarshal !!str `0.1` into config.RawService 
```

cdrage:
@ngtuna Any idea why this may be happening? I'm assuming problems with floating values not being converted.

sebgoa:
I am proposing a GSoC project to improve the dab support:
https://github.com/cncf/soc

cdrage:
@sebgoa Should we deprecate this feature since it's still experimental within Docker and at the moment doesn't work with Kompose?

sebgoa:
if it does not work, it is a regression that we should fix.

kadel:
Is DAB actually used somewhere? Do we know if someone is using dab even with docker?

sebgoa:
Honestly I have not kept up with the dab development. but I will check.

djkonro:
Please, I am interested in working on this project for the Google Summer of Code. I am very comfortable working with programming languages like Java, C/C++, Python and Go. I am very passionate about cloud and container technologies, and I would like to get some help that would enable me better familiarizing myself with the project. I am already familiar with Docker and Docker compose and I am currently familiarizing myself with Kubernetes.

cdrage:
I believe this link: https://github.com/docker/docker/issues/26876 would be the best in terms of keeping track *IF* DAB comes out of being experimental.

It *has* happened in the past that an experimental feature wasn't accepted into Docker. So it'd be best to keep track of this while we continue to support development / improvements.

cdrage:
I'm closing this for now until Docker figures out DAB's role.

PR #653, opened 2017-06-19
Disable dab

cdrage:
See issue: https://github.com/kubernetes-incubator/kompose/issues/390

Remove DAB as it is hard to maintain / not much usage / DAB is still
experimental in Docker and there hasn't been much movement:
https://github.com/moby/moby/issues/26876

MarkDeprecated does not work at the moment due to issue:
https://github.com/kubernetes-incubator/kompose/issues/652

However, that is not a blocker as we `fatalF` within `ValidateFlags`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -102,4 +102,8 @@ func init() {
 	RootCmd.PersistentFlags().StringArrayVarP(&GlobalFiles, "file", "f", []string{}, "Specify an alternative compose file")
 	RootCmd.PersistentFlags().StringVarP(&GlobalBundle, "bundle", "b", "", "Specify a Distributed Application Bundle (DAB) file")
 	RootCmd.PersistentFlags().StringVar(&GlobalProvider, "provider", "kubernetes", "Specify a provider. Kubernetes or OpenShift.")
+
+	// Mark DAB / bundle as deprecated, see issue: https://github.com/kubernetes-incubator/kompose/issues/390
+	// As DAB is still EXPERIMENTAL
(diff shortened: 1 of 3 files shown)


cdrage:
dabbing

ping @sebgoa @kadel

sebgoa:
so +1 , except that as you mention MarkDeprecated does not work.

cdrage:
@sebgoa Yes, an issue with Cobra, which already has a fix, see: https://github.com/spf13/cobra/pull/466 will just have to wait until it's merged upstream. But since we error out anyways if `--bundle` is passed, this will still work. We'll just have to update vendoring once https://github.com/spf13/cobra/pull/466 is merged :+1:

cdrage:
I've also updated the commit title / information. 

We're **disabling** DAB not removing it, as the code for conversion is still there for future use!

cdrage:
Okay. Tests pass now. @kadel @surajssd want to do a quick review?

surajnarwade:
works for me :)

procrypt:
@cdrage LGTM +1 for removing DAB ;)

procrypt:
@cdrage Do we need to write the `functional tests` for dab in `golang` since we are disabling it.

pradeepto:
> @cdrage Do we need to write the functional tests for dab in golang since we are disabling it.

No, we definitely shouldn't but I will let @cdrage and others take the final call.

cdrage:
@procrypt @surajnarwade Remeber to go through the review process and hit "approve" :+1: But since I got two confirmations, let's go ahead and merge this!

PR #660, opened 2017-06-20
Fix vet issues with build/push code

cdrage:
This fixes the vet issues when running `make vet` due to unkeyed fields.


cdrage:
Going to quickly merge this in since this is causing issues with those running local tests. Revert if this is incorrect!

PR #670, opened 2017-06-23
Add v3 example

cdrage:
Same example as docker-compose.yaml but with 3 in the version name.

Needed for 1.0.0 blog post!


kadel:
This is really bad example :-(
We shouldn't use it at all.

In most clusters it doesn't work without hacking it :-(

kadel:
Sorry, I'm completely wrong here. This one is OK

PR #684, opened 2017-07-06
Change menu to left side

cdrage:
This commit changes the menu to the left side rather than on top,
syncing with cdrage/minimal branch / style.


cdrage:
This syncs to the changes I made upstream at https://github.com/cdrage/minimal

It will now look like this:

image

surajnarwade:
this looks cool, LGTM :+1:

PR #685, opened 2017-07-07
Setup -> Installation + title updates

cdrage:
Updates the docs to reflect that it's not a "setup" documention but
rather an installation document.

I've also updated the header of architecture.md to reflect the filename.


surajnarwade:
LGTM :+1:

PR #690, opened 2017-07-07
Change menu to left side

cdrage:
This commit changes the menu to the left side rather than on top,
syncing with cdrage/minimal branch / style.


cdrage:
Mergin' after my screwups with the gh-pages branch.

PR #694, opened 2017-07-10
Fix redirection site issues

cdrage:
Fixes the issues in regards to rediction.

For example, supplying `/docs/conversion` within the user guide the link
will appear as a 404.

Adding this allows redirection from `/docs/conversion` to `/conversion`.


cdrage:
Doing a lazy merge as I'm doing some site changes / updates.

PR #700, opened 2017-07-11
Update mention bot config

cdrage:
Removes some of the values which are already set to their default values
on mention bot.


cdrage:
Going to merge this to fix the bot (wasn't correctly finding reviews due to it being labeled by the CLA bot before declared "open")

PR #702, opened 2017-07-12
remove unused parameter from ValidateComposeFile()

containscafeine:
In pkg/app/app.go, the function ValidateComposeFile() has an
unused parameter "cmd *cobra.Command".

This commit removes that parameter from the function and the
callers of the function.


procrypt:
LGTM

cdrage:
LGTM

PR #703, opened 2017-07-12
Refactoring code as per gosimple check

surajnarwade:
This PR refactors some code bits as per `gosimple` tool check.


cdrage:
LGTM!

PR #708, opened 2017-07-13
Conversion Reference -> Conversion Matrix

cdrage:
Initialize a documentation build (running the script on travis) as well
as change from Conversion Reference to Conversion Matrix.


cdrage:
Need to initiate a doc build (travis only) in order to effectively build this since incorporating recent changes, see: https://github.com/kubernetes/kompose/commits/gh-pages

PR #714, opened 2017-07-17
Moving version from variable to text file

### LINKED ISSUE #712: Kompose  latest release version info from version file
For fabrica8 maven plugin integration, we need a file which can give latest release version from Http URL.

Now this can be done by consuming github release API  but it will overkill the simplicity.
For the sake of consistency and aligning existing implementations we need the version file under kompose repository.
e.g. gofabric8 https://github.com/fabric8io/gofabric8/blob/master/version/VERSION

It would e great if kompose project can provide similar version file

surajnarwade:
Resolves #712 , As it is needed for fabric8-maven-plugin


surajnarwade:
cc: @surajssd for review

cdrage:
@surajnarwade 

I really don't like this and this over-complicates things. 

What happens if someone's developing in a different path?

What happens if someone doesn't have the source code?

What if GOPATH isn't set (they're just using Kompose on a production machine?)

```
version, _ := ioutil.ReadFile(path)
```

Is also not using `version, err` (you ignore it with `_`).


Let's not modify version.go and only add VERSION to `/build`, we can modify the `release.sh` script to make the variable substitution.

surajnarwade:
@cdrage got it, made changes accordingly

cdrage:
LGTM

PR #716, opened 2017-07-17
Update doc script

cdrage:
Redirection works for http://kompose.io/docs/conversion/ but not for
http://kompose.io/docs/conversion.md which is what I originally
intended.

This PR updates the script to redirect all `.md` doc links.


surajnarwade:
LGTM :)

PR #727, opened 2017-07-26
Add tests for deploy key specifically

cdrage:
Due to:
https://github.com/kubernetes/kompose/commit/9dcb2bfba6639c045022d732d49867689ebb1de9
tests for "replicas" was disabled due to generating "pod-only"
kubernetes artifacts as per "restart_policy" being added.

This PR adds "deploy" tests so we can effectively test `replicas`


cdrage:
I'm going to wait until #738 is merged in before merging this one in since this will create a conflict.

cdrage:
#738 is merged in. Let's wait for these tests to pass and then merge this in.

PR #731, opened 2017-07-28
Adds Kubernetes cluster tests

cdrage:
This adds cluster tests for "kompose up" and "kompose down" for
Kubernetes. At the moment this checks that they are deployable to a
single-node Kubernetes cluster.

More proficient tests such as testing if the pods are actually up will
be added in the future.

You can test this by running: `make test-ci` on your local-machine.

Furthermore, we'll eventually have this enabled on Fabric8 / CentOS CI /
Semaphore (whatever comes first) so we can have a full end-to-end
testing environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,4 +110,6 @@ test-image:
 test-container:
 	docker run -v `pwd`:/opt/tmp/kompose:ro -it $(TEST_IMAGE) 
 
-
+.PHONE: test-ci
+test-ci:
+	./script/test_ci/test.sh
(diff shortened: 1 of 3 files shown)


cdrage:
@surajnarwade @surajssd @ashetty1 @kadel 

Please review

ashetty1:
@cdrage when I run examples/docker-gitlab.yaml, it throws up volume mount related warnings; the pods are not brought up successfully. How about we add --emptyvols  for `kompose up` command?

cdrage:
@ashetty1 That kind of sucks that our current gitlab example doesn't work... I don't like how we have to pass in `--emptyvols`.. I think we should update our examples.

Yup, we are currently not checking if each pod are actually running post `kompose up`. I'll be adding that implementaiton later. Right now, it's getting `make test-ci` into Semaphore / some sort of CI.

Can you LGTM?

PR #736, opened 2017-08-01
Add integration doc

cdrage:
Adds an integration document that lists third-party organizations /
users that use Kompose in some-way or another.


docs/integrations.md
```diff
@@ -0,0 +1,33 @@
+# User Guide
```
surajnarwade: `Integrations` instead of `User guide`

cdrage: Woops. updated.

PR #737, opened 2017-08-01
Adds integrations doc

cdrage:
Adds integrations doc to website


cdrage:
#736 must be merged in first

PR #738, opened 2017-08-01
Add CPU limit, CPU Reservation and Memory Reservation

cdrage:
This adds support for CPU limit, CPU reservation as well as memory
reservation.

Specifically, when using the `deploy` key in Docker Compose.


surajnarwade:
LGTM

orkenstein:
The [doc](http://kompose.io/conversion/) says there's only support for `limit`. Are `reservations` supported?

PR #756, opened 2017-08-10
[testing ci] testing ci builds

cdrage:
testing to see if new builds successfully trigger all CI integrations
(semaphore, cla, travis, etc.)


cdrage:
Note to self: OpenShift tests are a bit finnicky at the moment due to low-memory on the VM. People may need to retry / rebuild some test cases.

PR #757, opened 2017-08-10
Update cluster tests, use local Kompose binary

cdrage:
Uses the local kompose binary that was built, not the one stored in the
GOPATH bin folder.

For example, using `./kompose` instead of `kompose`

Disables `make bin` generation for Kubernetes tests.


cdrage:
Hey @ashetty1 I'm going to merge this for now in order to make any new PR's pass for the cluster tests, let me know if this conflicts with anything.

PR #759, opened 2017-08-11
Adds healthcheck

cdrage:
This PR adds support for HealthCheck, being able to supply, for example:

```yaml
version: "3"

services:
  redis:
    image: redis
    healthcheck:
      test: echo "hello world"
      interval: 10s
      timeout: 1s
      retries: 5
```

Which is then converted to:

```yaml
spec:
  containers:
  - image: redis
    livenessProbe:
      exec:
        command:
        - echo "hello world"
      failureThreshold: 5
      periodSeconds: 10
      timeoutSeconds: 1
    name: redis
    resources: {}
  restartPolicy: Always
```

At the moment, this only supports livenessProbe, with support for readinessProbe in the future.

### DIFF
--- a/pkg/kobject/kobject.go
+++ b/pkg/kobject/kobject.go
@@ -99,8 +99,20 @@ type ServiceConfig struct {
 	Dockerfile      string              `compose:"dockerfile"`
 	Replicas        int                 `compose:"replicas"`
 	GroupAdd        []int64             `compose:"group_add"`
-	// Volumes is a struct which contains all information about each volume
-	Volumes []Volumes `compose:""`
+	Volumes         []Volumes           `compose:""`
+	HealthChecks    HealthCheck         `compose:""`
+}
+
+// HealthCheck the healthcheck configuration for a service
+// "StartPeriod" is not yet added to compose, see:
+// https://github.com/docker/cli/issues/116
+type HealthCheck struct {
+	Test        []string
+	Timeout     int32
+	Interval    int32
+	Retries     int32
+	StartPeriod int32
+	Disable     bool
 }
 
 // EnvVar holds the environment variable struct of a container
--- a/pkg/loader/compose/v3.go
+++ b/pkg/loader/compose/v3.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"strconv"
 	"strings"
+	"time"
 
 	libcomposeyaml "github.com/docker/libcompose/yaml"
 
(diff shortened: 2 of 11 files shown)


cdrage:
Note: Tests are still being added, but this is up for review :+1:

cdrage:
@surajssd @containscafeine @kadel @surajnarwade Can I get a review?

pkg/kobject/kobject.go
```diff
@@ -99,8 +98,20 @@ type ServiceConfig struct {
...
+
+// HealthCheck the healthcheck configuration for a service
+// "StartPeriod" is not yet added to compose, see:
+// https://github.com/docker/cli/issues/116
+type HealthCheck struct {
+	Test        []string
+	Timeout     int32
+	Interval    int32
+	Retries     int32
+	StartPeriod int32
```
containscafeine: Umm, should we add StartPeriod since it's not a part of upstream Compose, yet? I understand that soon it will be, but today this is invalid Compose syntax.
Should we just comment it out for now and enable it as soon as it's merged?
Also, an issue to track this would help.

pkg/loader/compose/v3.go
```diff
@@ -162,6 +163,53 @@ func loadV3Ports(ports []types.ServicePortConfig) []kobject.Ports {
...
+*/
+func parseHealthCheck(composeHealthCheck types.HealthCheckConfig) (kobject.HealthCheck, error) {
+
+	var timeout, interval, retries, startPeriod int32
+
+	// Here we convert the timeout from 1h30s (example) to 36030 seconds.
+	if composeHealthCheck.Timeout != "" {
+		parse, err := time.ParseDuration(composeHealthCheck.Timeout)
+		if err != nil {
+			return kobject.HealthCheck{}, errors.Wrap(err, "Unable to parse health check timeout variable")
```
containscafeine: nit: error message starts with a lowercase?
here and other places too?

pkg/loader/compose/v3.go
```diff
@@ -198,6 +246,20 @@ func dockerComposeToKomposeMapping(composeObject *types.Config) (kobject.Kompose
...
+		if composeServiceConfig.HealthCheck != nil {
+			var err error
+			serviceConfig.HealthChecks, err = parseHealthCheck(*composeServiceConfig.HealthCheck)
+			if err != nil {
+				return kobject.KomposeObject{}, errors.Wrap(err, "Unable to parse health check")
+			}
+		}
+
+		//Handling replicas
+		if composeServiceConfig.Deploy.Replicas != nil {
```
containscafeine: just curious, where did the replica bits come in the healthchecks PR?

pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -381,6 +381,33 @@ func (k *Kubernetes) UpdateKubernetesObjects(name string, service kobject.Servic
...
+				return errors.New("Health check must contain a command")
+			}
+
+			probe.TimeoutSeconds = service.HealthChecks.Timeout
+			probe.PeriodSeconds = service.HealthChecks.Interval
+			probe.FailureThreshold = service.HealthChecks.Retries
+
+			// See issue: https://github.com/docker/cli/issues/116
+			// StartPeriod has been added to docker/cli however, it is not yet added
+			// to compose. Once the feature has been implemented, this will automatically work
```
containscafeine: ack, but as of today it's invalid compose syntax, refer to comment above. WDYT?

pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -381,6 +381,33 @@ func (k *Kubernetes) UpdateKubernetesObjects(name string, service kobject.Servic
...
+			probe.InitialDelaySeconds = service.HealthChecks.StartPeriod
+
+			template.Spec.Containers[0].LivenessProbe = &probe
```
containscafeine: This is great, but how about generating both Liveness and Readiness probes from this, since in most of the cases both are the same. Just thinking out loud here, even I'm not very sure about this.

pkg/kobject/kobject.go
```diff
@@ -99,8 +98,20 @@ type ServiceConfig struct {
...
+	Disable     bool
```
containscafeine: Are we doing anything with this field in code, I couldn't find anything.
Should we not add a check that if this is false, make sure that no Kubernetes probes are generated?

pkg/kobject/kobject.go
cdrage: Basically, I have added it so that when it's finally added to Docker Compose python implementation as well as the docker CLI we don't have to do any work :) It'll automatically be there and working!

pkg/kobject/kobject.go
cdrage: Nope, I'll remove this.

pkg/loader/compose/v3.go
cdrage: Woops. Accidental copy and paste.

pkg/loader/compose/v3.go
cdrage: done.

pkg/transformer/kubernetes/k8sutils.go
cdrage: Since this is implemented, it will automatically work when the new key is added to Docker Compose

cdrage:
So essentially @containscafeine Health Check restarts the Docker container when it fails, LivenessProbe is the same within Kubernetes. Makes sense to only have LivenessProbe and add the ability to use Readiness when we add an external label :+1:

pkg/kobject/kobject.go
containscafeine: But `disable` is a valid field in Docker Compose, we should keep it. I think during transformation to Kubernetes objects we should add a check that if Disable is set to true, we should not generate any probes.

cdrage:
@containscafeine Added your suggestion regarding `Disable`!

cdrage:
Awesome!! @containscafeine I have rebased :+1:

PR #762, opened 2017-08-14
Go down to one ACK

cdrage:
Move down to one ACK rather than two for code changes. One or none for
doc changes.


cdrage:
ping @kadel @containscafeine @surajnarwade @surajssd

PR #766, opened 2017-08-16
[WIP] Update to use ./kompose instead of $GOPATH kompose binary

cdrage:
The issue is that when running `make test-cmd` you'll have to cp kompose
to the /usr/local/bin directory or the $GOPATH directory.

Since `make test` is being used consistently in the same git diretory,
it makes sense to use the local-built binary.


cdrage:
Blocked on #770 since `kompose version` does not fail if no kompose is located.

.travis.yml
```diff
@@ -4,9 +4,6 @@ sudo: required
 
 language: go
 
-# make this also work for forks
-go_import_path: github.com/kubernetes/kompose
```
kadel: Why is this removed? It shouldn't have any effect on this, or does it?

cdrage: Yeah, I need to re-work this...

cdrage:
Closing for now.

PR #767, opened 2017-08-16
Add compatibility matrix

cdrage:
Adds a matrix that signifies what versions of Kubernetes we support.


cdrage:
I'm going to update this at a later date with `Deployment` artifacts and such.

PR #771, opened 2017-08-16
Add deploy: mode: global support

cdrage:
Adds support for deploy: mode.

For example:

```yaml
version: "3"

services:
  foo:
    deploy:
      resources:
        mode: global
        replicas: 6
    image: redis
```

Will only generate replicas: 1 in Kubernetes pods as "global" limits
replicas to only one.

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -490,10 +490,17 @@ func (k *Kubernetes) ConfigEnvs(name string, service kobject.ServiceConfig) []ap
 func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.ServiceConfig, opt kobject.ConvertOptions) []runtime.Object {
 	var objects []runtime.Object
 	var replica int
+
 	if opt.IsReplicaSetFlag || service.Replicas == 0 {
 		replica = opt.Replicas
 	} else {
 		replica = service.Replicas
+	}
+
+	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
+	if service.DeployMode == "global" {
+		log.Warning("Global mode not yet supported, containers will only be replicated once throughout the cluster. DaemonSet support will be added in the future.")
+		replica = 1
 	}
 
 	if opt.CreateD {
--- a/script/test/cmd/tests.sh
+++ b/script/test/cmd/tests.sh
@@ -394,6 +394,15 @@ cd $CURRENT_DIR
 
 # Test V3 Support of Docker Compose
 
+# Test deploy mode: global
+cmd="kompose convert --stdout -j -f $KOMPOSE_ROOT/script/test/fixtures/v3/docker-compose-deploy-mode.yaml"
+sed -e "s;%VERSION%;$version;g" -e "s;%CMD%;$cmd;g"  "$KOMPOSE_ROOT/script/test/fixtures/v3/output-deploy-mode-k8s-template.json" > /tmp/output-k8s.json
+convert::expect_success_and_warning "kompose convert --stdout -j -f $KOMPOSE_ROOT/script/test/fixtures/v3/docker-compose-deploy-mode.yaml" "/tmp/output-k8s.json"
+
+cmd="kompose convert --stdout -j --provider=openshift -f $KOMPOSE_ROOT/script/test/fixtures/v3/docker-compose-deploy-mode.yaml"
+sed -e "s;%VERSION%;$version;g" -e "s;%CMD%;$cmd;g"  "$KOMPOSE_ROOT/script/test/fixtures/v3/output-deploy-mode-os-template.json" > /tmp/output-os.json
+convert::expect_success_and_warning "kompose convert --stdout -j --provider=openshift -f $KOMPOSE_ROOT/script/test/fixtures/v3/docker-compose-deploy-mode.yaml" "/tmp/output-os.json"
+
 # Test support for cpu and memory limits + reservations
 cmd="kompose convert --stdout -j -f $KOMPOSE_ROOT/script/test/fixtures/v3/docker-compose-memcpu.yaml"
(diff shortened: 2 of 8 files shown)


surajnarwade:
https://docs.docker.com/compose/compose-file/#mode says `global` spin up exactly one container per swarm *node*, can we achieve same with kubernetes ?

cdrage:
@surajnarwade nope, unless it's a daemonset. For now it's just `replicas: 1`. I really want to add DaemonSet however.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -496,6 +496,12 @@ func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.Servic
 		replica = service.Replicas
 	}
 
+	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
+	if service.DeployMode == "global" {
+		log.Infof("Deploy mode has been set to global. Setting Kubernetes container replicas to 0 for service %s", name)
```
kadel: Message says `replicas to 0` but it sets replicas to `1`

script/test/fixtures/v3/docker-compose-deploy-mode.yaml
```diff
@@ -0,0 +1,9 @@
+version: "3"
+
+services:
+  foo:
+    deploy:
+      resources:
```
kadel: `mode: ` is not under resource but directly under `deploy` key and the same thing for `replicas`

script/test/cmd/tests.sh
```diff
@@ -394,6 +394,15 @@ cd $CURRENT_DIR
 
 # Test V3 Support of Docker Compose
 
+# Test deploy mode: global
+cmd="kompose convert --stdout -j -f $KOMPOSE_ROOT/script/test/fixtures/v3/docker-compose-deploy-mode.yaml"
+sed -e "s;%VERSION%;$version;g" -e "s;%CMD%;$cmd;g"  "$KOMPOSE_ROOT/script/test/fixtures/v3/output-deploy-mode-k8s-template.json" > /tmp/output-k8s.json
+convert::expect_success "kompose convert --stdout -j -f $KOMPOSE_ROOT/script/test/fixtures/v3/docker-compose-deploy-mode.yaml" "/tmp/output-k8s.json"
```
kadel: use `convert::expect_success_and_warning` to also check warning to avoid false positives with the wrong docker compose file

cdrage:
Thanks @kadel I've gone ahead and implemented the changes.

script/test/fixtures/v3/docker-compose-deploy-mode.yaml
```diff
@@ -0,0 +1,9 @@
+version: "3"
+
+services:
+  foo:
+    deploy:
+      mode: global
+      resources:
+        replicas: 6
```
kadel: there is not `replicas` in the `resources`  field   ;-)

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -496,6 +496,12 @@ func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.Servic
 		replica = service.Replicas
 	}
 
+	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
+	if service.DeployMode == "global" {
+		log.Infof("Deploy mode has been set to global. Setting Kubernetes container replicas to 1 for service %s", name)
```
kadel: Isn't it a bit confusing to display this message even if replicas are set to 1 or  if `replicas` is not set?

cdrage: By default DeployMode is "replicated" so this will only be shown when "global" is selected. Should we remove this all-together?

script/test/fixtures/v3/docker-compose-deploy-mode.yaml
cdrage: Woops, again another error!

pkg/transformer/kubernetes/kubernetes.go
cdrage: Ah, I see what you mean now. Updated!

cdrage:
Again, another mistake and tests failed. But fixed with @kadel 's suggestions.

Ready for another review!

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -490,12 +490,19 @@ func (k *Kubernetes) ConfigEnvs(name string, service kobject.ServiceConfig) []ap
...
+
 	if opt.IsReplicaSetFlag || service.Replicas == 0 {
 		replica = opt.Replicas
 	} else {
 		replica = service.Replicas
 	}
 
+	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
+	if service.DeployMode == "global" {
+		log.Infof("Global mode has been set. Containers will only be replicated once throughout the cluster")
```
kadel: This message is still weird. It doesn't say what it does.
For kubernetes Kompose  is ignoring global mode, Kompose just creates regular Deployment with one replica.

So this is not what global mode should do. Should this message just say that global mode is not supported and that regular deployment will be created?

cdrage: @kadel In the future when we add DaemonSet we could switch it to use that, as that is better reflection on how this should work (one container per node). 

Would it be better to output a warning saying that Global is not support, but replicas will be set to 1?

kadel: > Would it be better to output a warning saying that Global is not support, but replicas will be set to 1?

Yes , I think this would be the best.

We can remove it once we start doing the DaemonSet thing. ( automatically using DaemonSets for services with global mode.)

cdrage: Luckily I doubt many people use `global`. So for now let's output the warning and put up an issue to eventually have it switch to daemonset

cdrage:
@kadel Updated with the new warning message

PR #783, opened 2017-08-21
add correct secrets for bintray

kadel:
this should fix `Bintray response: 401 Unauthorized. This resource requires authentication` error while uploading binaries to bintray

I'll add links to readme in another PR, once we are sure that this works.


cdrage:
LGTM

PR #787, opened 2017-08-21
Add --volumes parameter, deprecate emptyvols

cdrage:
This adds the --volumes paramater with a "generate" and "empty"

By default, "generate" will be used as a place-holder for "true".
Although not used in the code, we will eventually add "none"

This uses CLI paramater naming processes (no emptyVols as that is Go /
Kubernetes specific) and thus we use dashes.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -32,6 +32,7 @@ var (
 	ConvertBuildRepo             string
 	ConvertBuildBranch           string
 	ConvertBuild                 string
+	ConvertVolumes               string
 	ConvertChart                 bool
 	ConvertDeployment            bool
 	ConvertDaemonSet             bool
@@ -78,6 +79,7 @@ var convertCmd = &cobra.Command{
 			BuildBranch:                 ConvertBuildBranch,
 			CreateDeploymentConfig:      ConvertDeploymentConfig,
 			EmptyVols:                   ConvertEmptyVols,
+			Volumes:                     ConvertVolumes,
 			InsecureRepository:          ConvertInsecureRepo,
 			IsDeploymentFlag:            cmd.Flags().Lookup("deployment").Changed,
 			IsDaemonSetFlag:             cmd.Flags().Lookup("daemon-set").Changed,
@@ -128,9 +130,13 @@ func init() {
 	convertCmd.Flags().MarkShorthandDeprecated("y", "YAML is the default format now.")
 	convertCmd.Flags().BoolVarP(&ConvertJSON, "json", "j", false, "Generate resource files into JSON format")
 	convertCmd.Flags().BoolVar(&ConvertStdout, "stdout", false, "Print converted objects to stdout")
-	convertCmd.Flags().BoolVar(&ConvertEmptyVols, "emptyvols", false, "Use Empty Volumes. Do not generate PVCs")
 	convertCmd.Flags().StringVarP(&ConvertOut, "out", "o", "", "Specify a file name to save objects to")
 	convertCmd.Flags().IntVar(&ConvertReplicas, "replicas", 1, "Specify the number of repliaces in the generate resource spec")
+	convertCmd.Flags().StringVar(&ConvertVolumes, "volumes", "persistentVolumeClaim", `Volumes to be generated ("persistentVolumeClaim"|"emptyDir")`)
+
+	// Deprecated commands
+	convertCmd.Flags().BoolVar(&ConvertEmptyVols, "emptyvols", false, "Use Empty Volumes. Do not generate PVCs")
+	convertCmd.Flags().MarkDeprecated("emptyvols", "emptyvols has been marked as deprecated. Use --volumes empty")
 
 	// In order to 'separate' both OpenShift and Kubernetes only flags. A custom help page is created
 	customHelp := `Usage:{{if .Runnable}}
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -29,6 +29,7 @@ import (
 var (
 	UpReplicas     int
 	UpEmptyVols    bool
+	UpVolumes      string
 	UpInsecureRepo bool
 	UpNamespace    string
 	UpOpt          kobject.ConvertOptions
@@ -54,6 +55,7 @@ var upCmd = &cobra.Command{
 			InputFiles:         GlobalFiles,
 			Provider:           strings.ToLower(GlobalProvider),
 			EmptyVols:          UpEmptyVols,
+			Volumes:            UpVolumes,
 			Namespace:          UpNamespace,
 			InsecureRepository: UpInsecureRepo,
 			IsNamespaceFlag:    cmd.Flags().Lookup("namespace").Changed,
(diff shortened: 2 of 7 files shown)


ashetty1:
@cdrage I tried running this with etherpad docker-compose file, and it looks like `--volumes empty` option creates a pvc:

```
$ ./kompose up --provider=openshift  --volumes empty  -f script/test/fixtures/etherpad/docker-compose.yml 
WARN Unsupported depends_on key - ignoring        
INFO We are going to create OpenShift DeploymentConfigs, Services and PersistentVolumeClaims for your Dockerized application. 
 
INFO Deploying application in "myproject" namespace 
INFO Successfully created Service: etherpad       
INFO Successfully created Service: mariadb        
INFO Successfully created DeploymentConfig: etherpad 
INFO Successfully created ImageStream: etherpad   
INFO Successfully created DeploymentConfig: mariadb 
INFO Successfully created ImageStream: mariadb    
INFO Successfully created PersistentVolumeClaim: mariadb-claim0 of size 100Mi. If your cluster has dynamic storage provisioning, you don't have to do anything. Otherwise you have to create PersistentVolume to make PVC work 

Your application has been deployed to OpenShift. You can run 'oc get dc,svc,is,pvc' for details.

```

Let me know if I am doing anything wrong here.

cdrage:
@ashetty1 Odd! Thanks for noticing that, looks like I missed something in the `openshift.go` file.

cdrage:
@ashetty1 It works on my machine using `convert`...
```
▶ ./kompose convert --provider=openshift  --volumes empty  -f script/test/fixtures/etherpad/docker-compose.yml 
WARN Unsupported depends_on key - ignoring        
INFO OpenShift file "etherpad-service.yaml" created 
INFO OpenShift file "mariadb-service.yaml" created 
INFO OpenShift file "etherpad-deploymentconfig.yaml" created 
INFO OpenShift file "etherpad-imagestream.yaml" created 
INFO OpenShift file "mariadb-deploymentconfig.yaml" created 
INFO OpenShift file "mariadb-imagestream.yaml" created 
```

But not `up`.

Thanks for testing it! I'll go ahead and fix it.

cdrage:
@ashetty1 Okay, should be fixed now. I forgot to add the param to `up.go`. Please try again for me!

ashetty1:
Something's wrong here. @cdrage, can you do a fresh checkout and try?

```
$ ./kompose version
panic: up flag redefined: volumes

goroutine 1 [running]:
panic(0x19a4140, 0xc42049d590)
        /usr/lib/golang/src/runtime/panic.go:500 +0x1a1
github.com/kubernetes/kompose/vendor/github.com/spf13/pflag.(*FlagSet).AddFlag(0xc42034c3c0, 0xc42026cb40)
        /usr/lib/golang/src/github.com/kubernetes/kompose/vendor/github.com/spf13/pflag/flag.go:797 +0x9ea
github.com/kubernetes/kompose/vendor/github.com/spf13/pflag.(*FlagSet).VarPF(0xc42034c3c0, 0x29d8020, 0x2a2ba70, 0x1c9a1df, 0x7, 0x0, 0x0, 0x1cce0b1, 0x2c, 0xc42026caa0)
        /usr/lib/golang/src/github.com/kubernetes/kompose/vendor/github.com/spf13/pflag/flag.go:780 +0x141
github.com/kubernetes/kompose/vendor/github.com/spf13/pflag.(*FlagSet).VarP(0xc42034c3c0, 0x29d8020, 0x2a2ba70, 0x1c9a1df, 0x7, 0x0, 0x0, 0x1cce0b1, 0x2c)
        /usr/lib/golang/src/github.com/kubernetes/kompose/vendor/github.com/spf13/pflag/flag.go:786 +0x8e
github.com/kubernetes/kompose/vendor/github.com/spf13/pflag.(*FlagSet).StringVar(0xc42034c3c0, 0x2a2ba70, 0x1c9a1df, 0x7, 0x1c9af6b, 0x8, 0x1cce0b1, 0x2c)
        /usr/lib/golang/src/github.com/kubernetes/kompose/vendor/github.com/spf13/pflag/string.go:37 +0xa3
github.com/kubernetes/kompose/cmd.init.5()
        /usr/lib/golang/src/github.com/kubernetes/kompose/cmd/up.go:76 +0x1a9
github.com/kubernetes/kompose/cmd.init()
        /usr/lib/golang/src/github.com/kubernetes/kompose/cmd/version.go:46 +0xb5
main.init()
        /usr/lib/golang/src/github.com/kubernetes/kompose/main.go:24 +0x2e
```

cdrage:
@ashetty1 Ah, you're too fast! It didn't git push -f in time. Try now:

```
▶ ./kompose up --provider=openshift  --volumes empty  -f script/test/fixtures/etherpad/docker-compose.yml 
WARN Unsupported depends_on key - ignoring        
INFO We are going to create OpenShift DeploymentConfigs, Services and PersistentVolumeClaims for your Dockerized application. 
 
INFO Deploying application in "myproject" namespace 
INFO Successfully created Service: etherpad       
INFO Successfully created Service: mariadb        
INFO Successfully created DeploymentConfig: etherpad 
INFO Successfully created ImageStream: etherpad   
INFO Successfully created DeploymentConfig: mariadb 
INFO Successfully created ImageStream: mariadb    

Your application has been deployed to OpenShift. You can run 'oc get dc,svc,is,pvc' for details.
```

ashetty1:
Works now :)

cdrage:
@ashetty1 Awesome! Wanna do a quick review? :+1:

cdrage:
@ashetty1 I'll add some unit tests as well, could you do a code review for me?

cdrage:
@ashetty1 Unit test has been added!

kadel:
> This adds the --volumes paramater with a "generate" and "empty"
> 
> By default, "generate" will be used as a place-holder for "true".
> Although not used in the code, we will eventually add "none"


what does `generate` do?

cdrage:
@kadel Because if I set it to "none" and set that as default, it would seem as though no volumes are used / generated. Hence why I put "generate" and set that as a the default. Does that make sense?

kadel:
> @kadel Because if I set it to "none" and set that as default, it would seem as though no volumes are used / generated. Hence why I put "generate" and set that as a the default. Does that make sense?

I don't think that its good name :-(
It is not clear what it does :-(

I think that the best would be to use actual Kubernetes volume type names (https://kubernetes.io/docs/concepts/storage/volumes/#types-of-volumes).  It would be much cleaner .

The default should be `persistentVolumeClaim` - I know it is long :-( but it will be the default value so users don't have to write it.

Instead of empty I would put there the whole `emptyDir` and once implemented we will add `hostPath` 

what do you think?

cdrage:
@kadel To be honest I don't like using Kubernetes naming conventions in CLI tools. Not even `kubectl` uses that naming convention within their CLI. It's all lower-case with hyphens. 

Having the long names would make `kompose convert --help` output even longer. 

What about: `pvc | none | empty` for the options with `pvc` set as the default?

kadel:
> @kadel To be honest I don't like using Kubernetes naming conventions in CLI tools. Not even kubectl uses that naming convention within their CLI. It's all lower-case with hyphens.

I think this rule is only for flag name not for values.

> What about: pvc | none | empty for the options with pvc set as the default?

Problem with this is that it is still not clear what it does :-(

If we do `emptyDir`, `persistentVolumeClaim`, and `ignore` that it says everything. I don't think that few words more is a problem.

`emptyDir` is also easier to find in documentation that just `empty` same for `pvc` vs `persistentVolumeClaim`

cdrage:
@kadel Okay, that *would* make it easier.

For CLI params keep with the conventional way, but with values stick to Kubernetes.

For example, this this how they do --type in `kubectl`:

```sh
      --type='': Type for this service: ClusterIP, NodePort, or LoadBalancer. Default is 'ClusterIP'
```

I'll update this in this PR.

ashetty1:
@cdrage or should the value be just boolean? False/no would mean emptyvols; yes by default. WDYT

cdrage:
@ashetty1 A bit confused. --emptyvols is false by default, passing in --emptyvols by itself makes it true

cdrage:
@ashetty1 @kadel @surajnarwade @surajssd @containscafeine up for review

kadel:
We can't forget to put mentions about this to release notes when we do next release

cdrage:
@kadel Yup, will do! :+1:

PR #797, opened 2017-08-25
Fix kompose docker container wrong href

JadCham:
The link was pointing to komposeui instead of kompose-docker


cdrage:
@JadCham My mistake! Thanks :+1:

PR #802, opened 2017-08-30
Adding OpenShift tests for deploy keys

ashetty1:


cdrage:
LGTM, tests pass on Semaphore (with the exception of MariaDB CrashLoopBackOff...) and tests pass locally for me :+1:

PR #808, opened 2017-09-06
Remove redundant strings.ToLower()

cdrage:
Removes the redundant strings.ToLower commands for GlobalProvider in
up.go, convert.go and down.go


surajnarwade:
works for me, LGTM :+1:

PR #810, opened 2017-09-07
Added `--controller` feature

surajnarwade:
Previously we used to mention controller type as `--deployment`,
`--replication-controller` or `--daemonset` as argument.
But now,
it will be like,

ex.

```
kompose convert --controller=daemonset
```

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -17,6 +17,8 @@ limitations under the License.
 package cmd
 
 import (
+	"strings"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/kubernetes/kompose/pkg/app"
 	"github.com/kubernetes/kompose/pkg/kobject"
@@ -42,6 +44,7 @@ var (
 	ConvertInsecureRepo          bool
 	ConvertDeploymentConfig      bool
 	ConvertReplicas              int
+	ConvertController            string
 	ConvertOpt                   kobject.ConvertOptions
 )
 
@@ -78,6 +81,7 @@ var convertCmd = &cobra.Command{
 			IsDeploymentFlag:            cmd.Flags().Lookup("deployment").Changed,
 			IsDaemonSetFlag:             cmd.Flags().Lookup("daemon-set").Changed,
 			IsReplicationControllerFlag: cmd.Flags().Lookup("replication-controller").Changed,
+			Controller:                  strings.ToLower(ConvertController),
 			IsReplicaSetFlag:            cmd.Flags().Lookup("replicas").Changed,
 			IsDeploymentConfigFlag:      cmd.Flags().Lookup("deployment-config").Changed,
 		}
@@ -102,6 +106,10 @@ func init() {
 	convertCmd.Flags().BoolVar(&ConvertDaemonSet, "daemon-set", false, "Generate a Kubernetes daemonset object")
 	convertCmd.Flags().BoolVarP(&ConvertDeployment, "deployment", "d", false, "Generate a Kubernetes deployment object")
 	convertCmd.Flags().BoolVar(&ConvertReplicationController, "replication-controller", false, "Generate a Kubernetes replication controller object")
+	convertCmd.Flags().StringVar(&ConvertController, "controller", "", `Set the output controller ("deployment"|"daemonSet"|"replicationController")`)
+	convertCmd.Flags().MarkDeprecated("daemon-set", "use --controller")
+	convertCmd.Flags().MarkDeprecated("deployment", "use --controller")
+	convertCmd.Flags().MarkDeprecated("replication-controller", "use --controller")
 	convertCmd.Flags().MarkHidden("chart")
 	convertCmd.Flags().MarkHidden("daemon-set")
 	convertCmd.Flags().MarkHidden("replication-controller")
@@ -112,6 +120,7 @@ func init() {
 	convertCmd.Flags().BoolVar(&ConvertInsecureRepo, "insecure-repository", false, "Use an insecure Docker repository for OpenShift ImageStream")
 	convertCmd.Flags().StringVar(&ConvertBuildRepo, "build-repo", "", "Specify source repository for buildconfig (default remote origin)")
 	convertCmd.Flags().StringVar(&ConvertBuildBranch, "build-branch", "", "Specify repository branch to use for buildconfig (default master)")
+	convertCmd.Flags().MarkDeprecated("deployment-config", "use --controller")
 	convertCmd.Flags().MarkHidden("deployment-config")
 	convertCmd.Flags().MarkHidden("insecure-repository")
 	convertCmd.Flags().MarkHidden("build-repo")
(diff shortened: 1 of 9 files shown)


cdrage:
@surajnarwade is this still WIP?

I don't see the deprecation of the current commands (`--daemonset`), I also see removals of `IsDaemonSetFlag` which is not good since it's used for the `--daemonset` command...

cmd/convert.go
```diff
@@ -62,28 +60,24 @@ var convertCmd = &cobra.Command{
...
+			Build:                  ConvertBuild,
+			BuildRepo:              ConvertBuildRepo,
+			BuildBranch:            ConvertBuildBranch,
+			CreateDeploymentConfig: ConvertDeploymentConfig,
+			EmptyVols:              ConvertEmptyVols,
+			InsecureRepository:     ConvertInsecureRepo,
+			Controller:             strings.ToLower(ConvertController),
+			IsReplicaSetFlag:       cmd.Flags().Lookup("replicas").Changed,
+			IsDeploymentConfigFlag: cmd.Flags().Lookup("deployment-config").Changed,
+			IsDeploymentFlag:       cmd.Flags().Lookup("controller").Changed,
```
cdrage: You are removing all the other `Is..` flags :-1:

pkg/app/app.go
```diff
@@ -156,28 +150,7 @@ func validateControllers(opt *kobject.ConvertOptions) {
...
-			}
-			if opt.CreateRC {
-				count++
-			}
-			if count > 1 {
-				log.Fatalf("Error: only one kind of Kubernetes resource can be generated when --out or --stdout is specified")
-			}
-		}
-
-	} else if opt.Provider == "openshift" {
```
cdrage: Why are you removing all of this?  Where is the replacement code?

pkg/app/app.go
```diff
@@ -72,24 +72,18 @@ func ValidateFlags(bundle string, args []string, cmd *cobra.Command, opt *kobjec
...
-		if daemonSet {
-			log.Fatalf("--daemon-set is a Kubernetes only flag")
-		}
-		if replicationController {
-			log.Fatalf("--replication-controller is a Kubernetes only flag")
-		}
-		if deployment {
-			log.Fatalf("--deployment, -d is a Kubernetes only flag")
+		if controller == "daemonset" || controller == "replicationcontroller" {
+			log.Fatalf("--controller is a Kubernetes only flag")
```
cdrage: No it's not a Kubernetes only flag :-1: `--controller` is meant to be added for OpenShift as well..

pkg/kobject/kobject.go
```diff
@@ -32,30 +32,28 @@ type KomposeObject struct {
...
+	InsecureRepository     bool
+	Replicas               int
+	InputFiles             []string
+	OutFile                string
+	Provider               string
+	Namespace              string
+	Controller             string
+	IsDeploymentFlag       bool
+	//IsDaemonSetFlag             bool
+	//IsReplicationControllerFlag bool
```
cdrage: Why commented out? We are still going to be using these for the deprecated options.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -502,14 +502,15 @@ func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.Servic
...
+	if opt.Controller == "deployment" && opt.Provider == "kubernetes" {
+		//fmt.Println("created")
 		objects = append(objects, k.InitD(name, service, replica))
 	}
-	if opt.CreateDS {
+	if opt.Controller == "daemonset" {
 		objects = append(objects, k.InitDS(name, service))
 	}
-	if opt.CreateRC {
+	if opt.Controller == "replicationcontroller" {
```
cdrage: you are checking against `replicationcontroller`. Whatever is coming from `opt.Controller` should be lowercased so that options such as `replicationController` work.

cdrage:
@surajnarwade awesome work, much better! :+1: I'll test this out when I have some time.

surajnarwade:
issue is being tracked here https://github.com/kubernetes/kompose/issues/792

cdrage:
The code LGTM but tests need to be added (again, don't change any old ones / current ones, just add new tests) @surajnarwade

surajnarwade:
@cdrage , added tests

cmd/convert.go
```diff
@@ -100,8 +104,12 @@ func init() {
 	// Kubernetes only
 	convertCmd.Flags().BoolVarP(&ConvertChart, "chart", "c", false, "Create a Helm chart for converted objects")
 	convertCmd.Flags().BoolVar(&ConvertDaemonSet, "daemon-set", false, "Generate a Kubernetes daemonset object")
-	convertCmd.Flags().BoolVarP(&ConvertDeployment, "deployment", "d", false, "Generate a Kubernetes deployment object")
+	convertCmd.Flags().BoolVar(&ConvertDeployment, "deployment", false, "Generate a Kubernetes deployment object")
```
cdrage: missed this, but please leave this unmodified, since we are doing backwards compatibility now, leave the `d` in.

cmd/convert.go
```diff
@@ -100,8 +104,12 @@ func init() {
...
 	convertCmd.Flags().BoolVar(&ConvertReplicationController, "replication-controller", false, "Generate a Kubernetes replication controller object")
+	convertCmd.Flags().StringVar(&ConvertController, "controller", "", `Set the type of controller ("deployment"|"daemonSet"|"replicationController")`)
```
cdrage: Maybe change to `Set the output controller` instead.

pkg/app/app.go
```diff
@@ -76,6 +76,10 @@ func ValidateFlags(bundle string, args []string, cmd *cobra.Command, opt *kobjec
 	replicationController := cmd.Flags().Lookup("replication-controller").Changed
 	deployment := cmd.Flags().Lookup("deployment").Changed
 
+	// Get the controller
+	controller := cmd.Flags().Lookup("controller").Value.String()
```
cdrage: I thought I put my comment on this, but no need to do a look up. Just use `opt`. `controller := opt.Controller`

pkg/app/app.go
```diff
@@ -101,6 +108,9 @@ func ValidateFlags(bundle string, args []string, cmd *cobra.Command, opt *kobjec
 		if buildBranch {
 			log.Fatalf("--build-branch is an Openshift only flag")
 		}
+		if controller == "deploymentconfig" {
+			log.Fatalf("--controller=deploymentconfig is OpenShift only flag")
```
cdrage: `is an Openshift only flag`

surajnarwade:
@cdrage review needed here :)

pkg/app/app.go
```diff
@@ -101,6 +108,9 @@ func ValidateFlags(bundle string, args []string, cmd *cobra.Command, opt *kobjec
 		if buildBranch {
 			log.Fatalf("--build-branch is an Openshift only flag")
 		}
+		if controller == "deploymentconfig" {
+			log.Fatalf("--controller=deploymentconfig is an OpenShift only flag")
```
cdrage: --controller=deploymentConfig

surajnarwade:
@cdrage , fixed changes :) ready to merge

cdrage:
LGTM

PR #811, opened 2017-09-11
emptvols -> volumes=yes

ashetty1:


cdrage:
LGTM

surajnarwade:
LGTM :)

script/test_in_openshift/lib.sh
```diff
@@ -192,7 +192,7 @@ function convert::kompose_up () {
     # Usage: convert::kompose_up <docker_compose_file>
     local compose_file=$1
     convert::print_msg "Running kompose up ..."
-    ./kompose up --provider=openshift --emptyvols -f $compose_file
+    ./kompose up --provider=openshift --volumes empty -f $compose_file
```
kadel: This is suspicious. 
Isn't correct value `emptyDir`? It looks like there is no validation for `--volumes` values.
ping @cdrage

kadel: created https://github.com/kubernetes/kompose/issues/814

ashetty1: @kadel looking into it. You are right. Also, the help text needs to change,

PR #813, opened 2017-09-12
Added feature for `placement` key in v3

surajnarwade:
it will map `engine.labels.operatingsystem` to `beta.kubernetes.io/os` and
`node.hostname` to `kubernetes.io/hostname` and all other constraints will not be supported.

### DIFF
--- a/pkg/loader/compose/v3.go
+++ b/pkg/loader/compose/v3.go
@@ -294,6 +294,20 @@ func dockerComposeToKomposeMapping(composeObject *types.Config) (kobject.Kompose
 			serviceConfig.Replicas = int(*composeServiceConfig.Deploy.Replicas)
 		}
 
+		// placement:
+		placement := make(map[string]string)
+		for _, j := range composeServiceConfig.Deploy.Placement.Constraints {
+			p := strings.Split(j, " == ")
+			if p[0] == "node.hostname" {
+				placement["kubernetes.io/hostname"] = p[1]
+			} else if p[0] == "engine.labels.operatingsystem" {
+				placement["beta.kubernetes.io/os"] = p[1]
+			} else {
+				log.Warn(p[0], " constraints in placement is not supported, only 'node.hostname' and 'engine.labels.operatingsystem' is only supported as a constraint ")
+			}
+		}
+		serviceConfig.Placement = placement
+
 		// TODO: Build is not yet supported, see:
 		// https://github.com/docker/cli/blob/master/cli/compose/types/types.go#L9
 		// We will have to *manually* add this / parse.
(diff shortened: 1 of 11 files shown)


cdrage:
@surajnarwade Code looks good, but can you add some integration / cmd tests? Then LGTM from me :+1:

surajnarwade:
@cdrage , tests added :)

cdrage:
One last thing, can you update `docs/conversion.md`? 

Otherwise, code LGTM. Once docs are updated let's merge.

surajnarwade:
@cdrage updated `docs/conversion.md`

pkg/loader/compose/v3.go
```diff
@@ -294,6 +294,20 @@ func dockerComposeToKomposeMapping(composeObject *types.Config) (kobject.Kompose
 			serviceConfig.Replicas = int(*composeServiceConfig.Deploy.Replicas)
 		}
 
+		// placement:
+		placement := make(map[string]string)
+		for _, j := range composeServiceConfig.Deploy.Placement.Constraints {
+			p := strings.Split(j, " == ")
```
cdrage: missed this, but it's a pain in the butt, but can you test for:
```
      placement:
        constraints: [node.role == manager]
```

as well, as per: https://docs.docker.com/compose/compose-file/#replicas

surajnarwade: @cdrage , added functional test for the same example and it's working

cdrage:
@surajnarwade Alright, code LGTM (thanks for adding another functional test).

Fix conflicts and then it should be good to go.

surajnarwade:
@cdrage fixed conflicts

pkg/loader/compose/v3.go
```diff
@@ -294,6 +294,20 @@ func dockerComposeToKomposeMapping(composeObject *types.Config) (kobject.Kompose
...
+			if p[0] == "node.hostname" {
+				placement["kubernetes.io/hostname"] = p[1]
+			} else if p[0] == "engine.labels.operatingsystem" {
+				placement["beta.kubernetes.io/os"] = p[1]
+			} else {
+				log.Warn(p[0], " constraints in placement is not supported")
```
cdrage: So just a second review, but maybe we should add that only `node.hostname` and `engine.labels.operatingsystem` is only supported as a constraint in the warning.

script/test/cmd/tests.sh
```diff
@@ -574,5 +597,7 @@ cmd="kompose convert --provider=openshift --stdout -j -f $KOMPOSE_ROOT/examples/
 sed -e "s;%VERSION%;$version;g" -e "s;%CMD%;$cmd;g"  $KOMPOSE_ROOT/script/test/fixtures/examples/output-gitlab-os.json > /tmp/output-os.json
 convert::expect_success_and_warning "kompose convert --provider=openshift --stdout -j -f $KOMPOSE_ROOT/examples/docker-gitlab.yaml" "/tmp/output-os.json"
 
+
+
```
cdrage: remove extra newlines

cdrage:
LGTM

PR #816, opened 2017-09-14
Fixed tmpfs with mode failure

### LINKED ISSUE #807: tmpfs with mode fails
Hello, 

Trying to use kompose on [microservices-demo](https://github.com/microservices-demo/microservices-demo), I had to modify the generated files, because a mode is provided for the tmpfs. 
Here is a shortened version of it:

Using this compose file. 

```yml
version: "2"

services:

  redis-master:
    image: gcr.io/google_containers/redis:e2e 
    ports:
      - "6379"
    tmpfs:
      - /tmp:rw,noexec,nosuid
```

This file can be used correctly with docker compose. However, when running `kompose up`, the pod fails to be created: 
```
redis-master-1539052294-z8txm   0/1       rpc error: code = 2 desc = Error response from daemon: {"message":"invalid bind mount spec \"/var/lib/kubelet/pods/c902b3c2-92d6-11e7-8f5f-0200ac100026/volumes/kubernetes.io~empty-dir/redis-master-tmpfs0:/tmp:rw,noexec,nosuid\": invalid mode: rw,noexec,nosuid"}   0         47s

```

```bash 
kompose version
1.1.0 (36652f6)
```

surajnarwade:
Fixes #807
now kompose will ignore mode of tmpfs and will pass only mount path.

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -45,11 +45,12 @@ import (
 	"k8s.io/kubernetes/pkg/runtime"
 	"k8s.io/kubernetes/pkg/util/intstr"
 	//"k8s.io/kubernetes/pkg/controller/daemon"
+	"sort"
+	"strings"
+
 	"github.com/pkg/errors"
 	"k8s.io/kubernetes/pkg/api/meta"
 	"k8s.io/kubernetes/pkg/labels"
-	"sort"
-	"strings"
 )
 
 // Kubernetes implements Transformer interface and represents Kubernetes transformer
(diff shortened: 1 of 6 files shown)


pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -351,7 +352,8 @@ func (k *Kubernetes) ConfigTmpfs(name string, service kobject.ServiceConfig) ([]
 	for index, volume := range service.TmpFs {
 		//naming volumes if multiple tmpfs are provided
 		volumeName := fmt.Sprintf("%s-tmpfs%d", name, index)
-
+		v := strings.Split(volume, ":")
+		volume = v[0]
```
cdrage: this could be just `volume = strings.Split(volume, ":")` remove the other line

surajnarwade: No, it will not work, 

```
# github.com/kubernetes/kompose/pkg/transformer/kubernetes
pkg/transformer/kubernetes/kubernetes.go:355: cannot use strings.Split(volume, ":") (type []string) as type string in assignment
```

cdrage: @surajnarwade Pretty easy to figure it out, I forgot to add the [0].
Do: `strings.Split(volume, ":")[0]`

surajnarwade:
issue is being tracked here https://github.com/kubernetes/kompose/issues/807

cdrage:
@surajnarwade can you add tests?

After that, LGTM.

surajnarwade:
added functional test

cdrage:
LGTM! Thanks for adding the tests.

PR #818, opened 2017-09-15
Simplify copy/paste installation for Linux

abitrolly:


cdrage:
Hey @abitrolly 

We are following the convention as shown by https://kubernetes.io/docs/tasks/tools/install-kubectl/ where we separate the moving of files as well as chmod +x in order to emphasize what commands are needed (and to prevent people copying+pasting and messing something up on their system).

abitrolly:
Make sense. I usually merge this into a single shell block, like this:
```
curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl

# Make the kubectl binary executable.
chmod +x ./kubectl

# Move the binary in to your PATH.
sudo mv ./kubectl /usr/local/bin/kubectl
```

Issue #859, opened 2017-11-06
Docker compose links section can't support


xianlubird:
About compose  introduce links https://docs.docker.com/compose/compose-file/compose-file-v2/#links .
  I have a compose like this:
```
web:
  image: registry.aliyuncs.com/acs-sample/wordpress:4.5
  ports:
    - '80'
  environment:
    WORDPRESS_AUTH_KEY: changeme
    WORDPRESS_SECURE_AUTH_KEY: changeme
    WORDPRESS_LOGGED_IN_KEY: changeme
    WORDPRESS_NONCE_KEY: changeme
    WORDPRESS_AUTH_SALT: changeme
    WORDPRESS_SECURE_AUTH_SALT: changeme
    WORDPRESS_LOGGED_IN_SALT: changeme
    WORDPRESS_NONCE_SALT: changeme
    WORDPRESS_NONCE_AA: changeme
  restart: always
  links:
    - 'db:mysql'
db:
  image: registry.aliyuncs.com/acs-sample/mysql:5.7
  environment:
    MYSQL_ROOT_PASSWORD: password
  restart: always
  labels:
    aliyun.logs: /var/log/mysql
```
  kompose will convert db service to db deployment, but the wordpress deplotment will connect db by `mysql` hostname.   So wordpress pods will call that and it can't connect the db database.  Because the `links` can't convert to k8s yaml.

  I have two kinds suggestions:
1.  kompose convert db service to mysql service, just rename deployment name for links, and the new service name is mysql, the wordpress can connect mysql by service name corrently.
2. print warning or error message if compose has links section which alias name is different from service name.

   Can anyone give me some suggestion, and I will create a PR to fix this, Thanks.

surajnarwade:
I think this will be bit confusing since name of deployment and service will change,

> kompose convert db service to mysql service, just rename deployment name for links, and the new service name is mysql, the wordpress can connect mysql by service name corrently.

This second option, I think will be better,

> print warning or error message if compose has links section which alias name is different from service name.

@cdrage thought here

cdrage:
Yeah, we should add the second option since the `links` key isn't able to be effectively converted to Kubernetes. We should clarify more about *why* it doesn't work and what the alternatives are.

Thanks @xianlubird for pointing this out!

PR #862, opened 2017-11-07
Ignore links attribute and print warning message

xianlubird:
Related to issue    #859 

Signed-off-by: xianlubird@gmail.com


xianlubird:
Done

surajnarwade:
@xianlubird , please fix functional test examples accordingly

xianlubird:
@surajnarwade Done, pls review, thanks.

pkg/loader/compose/compose.go
```diff
@@ -120,6 +121,28 @@ func checkUnsupportedKey(composeProject *project.Project) []string {
 							yamlTagName = "networks"
 						}
 					}
+
+					if f.Name() == "Links" {
+						//Links has "SERVICE:ALIAS" style, we don't support SERVICE != ALIAS
+						if linksArray := val.FieldByName(f.Name()); linksArray.Kind() == reflect.Slice {
```
cdrage: nested if, may be better to do `if f.Name() == "Links" && if linksArray := val.FieldByName(f.Name()); linksArray.Kind() == reflect.Slice {`

pkg/loader/compose/compose.go
```diff
@@ -120,6 +121,28 @@ func checkUnsupportedKey(composeProject *project.Project) []string {
...
+								if tmpLink := linksArray.Index(i); tmpLink.Kind() == reflect.String {
+									tmpLinkStr := tmpLink.String()
+									tmpLinkStrSplit := strings.Split(tmpLinkStr, ":")
+									if len(tmpLinkStrSplit) == 2 && tmpLinkStrSplit[0] != tmpLinkStrSplit[1] {
+										findUnsupportedLinksFlag = true
+										break
+									}
+								}
+							}
+							if !findUnsupportedLinksFlag {
```
cdrage: Looking at the code, I don't think this is doing anything? Since `findUnsupportedLinksFlag` is outside the `for i := 0; i < linksArray.Len(); i++ {` loop. I could be wrong!

pkg/loader/compose/compose.go
xianlubird: Done

pkg/loader/compose/compose.go
xianlubird: This flag `findUnsupportedLinksFlag ` is for the above code `for _, f := range s.Fields() {` at about 100 lines this file.The flag's goal is to find compose links like 
```
  links:
    - 'db:mysql'
```
which service name is different from alias name.   If we find links like this, this flag is set to `true` and code will process, `keysFound = append(keysFound, yamlTagName)` this code will append `links` to `keysFound` and print warning messsage. But if we have links like this
```
  links:                                links:
    - redis                              - 'mysql:mysql'
```
There are many compose like this in our examples and test case folder. For this kind of compose ,we can convert it and shouldn't print any warning message. So in this situation, `findUnsupportedLinksFlag` will set to `false`, and `links` will not be append to `keysFound`.  This code `keysFound = append(keysFound, yamlTagName)` won't run because we have continue, and code will process next filed.

surajnarwade:
LGTM

pkg/loader/compose/compose.go
cdrage: Ah, that makes sense. LGTM.

xianlubird:
Thanks for @cdrage @surajnarwade review.    So when will this PR be merged ?

PR #868, opened 2017-11-14
Adding `--controller` flag in `up` & `down`

### LINKED ISSUE #798: Up and convert commands are not equal in features
For example, we are able to deploy --daemon-set using `convert`, but not with `up`.

We have *some* things such as `--build` which is consistent in `kompose up`, but other aspects such as controllers, etc do not function correctly.

We should have the SAME experience between `up` and `convert`.

It goes against UX/UI having inconsistencies between the commands. One should convert, one should bring those converted up.

List of parameters NOT in `up`:
- [ ] --daemon-set
- [ ] --deployment
- [ ] --chart (I don't think this should be applicable)
- [ ] --replication-controller
- [x] --build-branch
- [x] --build-repo
- [ ] --deployment-config
- [x] --insecure-repository

A couple of these parameters will be deprecated in favour of `--controller` however.

surajnarwade:
To make `kompose up` & `kompose convert` equal in feature, This PR will
add `--controller` flag for `kompose up` as well as `kompose down`
so that user experience will be the same for `up` & `convert`
Resolves #798

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -805,6 +805,21 @@ func (k *Kubernetes) Deploy(komposeObject kobject.KomposeObject, opt kobject.Con
 				return err
 			}
 			log.Infof("Successfully created Deployment: %s", t.Name)
+
+		case *extensions.DaemonSet:
+			_, err := client.DaemonSets(namespace).Create(t)
+			if err != nil {
+				return err
+			}
+			log.Infof("Successfully created DaemonSet: %s", t.Name)
+
+		case *api.ReplicationController:
+			_, err := client.ReplicationControllers(namespace).Create(t)
+			if err != nil {
+				return err
+			}
+			log.Infof("Successfully created ReplicationController: %s", t.Name)
+
 		case *api.Service:
 			_, err := client.Services(namespace).Create(t)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


cdrage:
@surajnarwade Need tests :+1:

surajnarwade:
@cdrage since this feature is regarding `kompose up` and `kompose down`, there will be integration tests.
cc @ashetty1

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -797,6 +797,21 @@ func (k *Kubernetes) Deploy(komposeObject kobject.KomposeObject, opt kobject.Con
 				return err
 			}
 			log.Infof("Successfully created Deployment: %s", t.Name)
+
+		case *extensions.DaemonSet:
```
cdrage: Not too sure why this code is being added? Can you describe this at least within the commit description?

surajnarwade: Sure I will mention

surajnarwade: @cdrage , updated commit message

cdrage: please update GitHub description too...

surajnarwade:
@cdrage , review needed :)

script/test_k8s/kubernetes.sh
```diff
@@ -103,6 +103,19 @@ test_k8s() {
     sleep 2 # Sleep for k8s to catch up to deployment
     echo -e "\n${RED}kompose down -f $f ${NC}\n"
     ./kompose down -f $f
+    echo -e "\nTesting controller=daemonset key\n"
```
cdrage: I think we should find an alternative way to write these integration tests (perhaps in a separate function? and only running the httpd.yaml test?) or even a separate file...

It's a bit messy at the moment and we really need to refactor everything (using `e2e_test.go` similar to our other project we're working on.

Do you mind opening an issue explaining that we should refactor our integration tests to use a Go file / make this better? As well as add a comment (and spacing above this line) with a `# TODO: need to reorganize / refactor`
 Otherwise, this LGTM (for now).

surajnarwade: Thanks for adding issue as well.
Ready to merge ?

cdrage:
LGTM! Let's merge!

PR #882, opened 2017-11-30
Test cross-compiling within travis

cdrage:
Tests cross compiling by building on 1.6 of Go for Windows / Linux /
macOS


cdrage:
Merging just so we can get builds working. Please revert if there's an issue :+1:

PR #894, opened 2017-12-15
[RFC] Map ingress path to optional 'kompose.service.expose_path'

wms:
I've cobbled this together to allow a new optional label `kompose.service.expose_path` to map to `ingress.spec.rules[].http.paths[].path`.

This supports a workflow I find desirable, and I suspect others may find this useful too. I've never worked with Go before, so please forgive me if the implementation is sub-optimal. Also, I'm aware that you may not want to maintain such a specific feature, so I didn't spend any time adding tests or docs at this point.

If a maintainer here feels the concept is sound but has implementation advice, I'm more than happy to spend some time making this merge-worthy.

### DIFF
--- a/pkg/kobject/kobject.go
+++ b/pkg/kobject/kobject.go
@@ -92,6 +92,7 @@ type ServiceConfig struct {
 	Build           string              `compose:"build"`
 	BuildArgs       map[string]*string  `compose:"build-args"`
 	ExposeService   string              `compose:"kompose.service.expose"`
+	ExposePath      string              `compose:"kompose.service.expose_path"`
 	Stdin           bool                `compose:"stdin_open"`
 	Tty             bool                `compose:"tty"`
 	MemLimit        yaml.MemStringorInt `compose:"mem_limit"`
--- a/pkg/loader/compose/v1v2.go
+++ b/pkg/loader/compose/v1v2.go
@@ -241,6 +241,8 @@ func libComposeToKomposeMapping(composeObject *project.Project) (kobject.Kompose
 				serviceConfig.ServiceType = serviceType
 			case "kompose.service.expose":
 				serviceConfig.ExposeService = strings.ToLower(value)
+			case "kompose.service.expose_path":
+				serviceConfig.ExposePath = value
 			}
 		}
 		err = checkLabelsPorts(len(serviceConfig.Port), composeServiceConfig.Labels["kompose.service.type"], name)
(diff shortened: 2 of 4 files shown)


warrenseymour:
CLA signed

Code0x58:
When you do get a thumbs up and move onto the documentation+tests, you may find #896 helpful (probably just the paths) as it does the same kind of thing. It would also be good to return an error if tag is used without `kompose.service.expose`.

_p.s. I prefer the idea of `kompose.service.expose.path`_

cdrage:
@kadel @surajnarwade `kompose.service.expose.path` or `kompose.service.expose_path`, thoughts? I'm leaning towards `kompose.service.expose.path`

Tests / documentation should be added later, but we can help with that.

kadel:
If we are going with `kompose.service.expose.tls-secret` in #896 than `kompose.service.expose.path` makes much more sense here.

This PR goes in the right direction, it just needs to check for the case where `kompose.service.expose.path` is set but  `kompose.service.expose` is not (or it is false).

cdrage:
Hey @warrenseymour ! Please see @kadel 's comment. Once updated, I'd be good for merging. `kompose.expose.path` works.

hangyan:
@cdrage Is there any way to make this PR continue?

cdrage:
@warrenseymour Could we possibly continue using your code? :+1: Just seeing if you're still interested in completing this :smile:

josipradic:
I couldn't find any implementations regarding this subject, do you have plans for supporting this feature? We would find it quite useful for our CICD integration.

hangyan:
port

hangyan:
I can continue to work on this feature.

hangyan:
@ josipradic  can you add more details about your requirement. Since `kompose.service.expose` support multiple hostnames, so there will be multiple rules. if we support specific path, how would that apply? Any suggestions?

josipradic:
> @ josipradic can you add more details about your requirement. Since `kompose.service.expose` support multiple hostnames, so there will be multiple rules. if we support specific path, how would that apply? Any suggestions?

Hi @hangyan, maybe instead of adding support of one more label such as `kompose.service.expose.path(s)`, the existing one `kompose.service.expose` could be accepting hostnames as it is accepting now with the exception that if it includes `/{path}` to apply that path on that hostname? Without the `/{path}` addition it would work as it works now.

hangyan:
@josipradic  This is a good idea. I will create a patch to fix this.

PR #896, opened 2017-12-17
Add kompose.service.expose.tls-secret

Code0x58:


cdrage:
Tests pass (despite us having a hard barrier-to-entry to creating them, haha), code looks great, name makes sense. This LGTM. :+1: thanks a lot @Code0x58 !

Code0x58:
Thanks, I definitely made use of CI on here while working out the tests - sempahore shows 12 builds.

I think ingress TLS secrets won't map nicely between docker-compose/swarm and Kubernetes as it would be an application concern in compose/swam, so it will probably always be stuck in labels/workarounds like this. Pod `ImagePullSecret`s are probably in the same boat.

I image that down the line when compose secrets are supported, you could specify your secret within the compose file, then refer to it in this label without any additional work - so this should be a stable approach.

cdrage:
Yeah, there's absolutely no way to map it correctly without doing something *really* hacky. I agree that labels are the right way to go. 

This LGTM :+1: Thanks for clarifying regarding the secret and TLS. 

IMO this should have another review. @kadel @containscafeine @surajnarwade can you have a look?

cdrage:
Mergin' thanks @Code0x58 !

Issue #883, opened 2017-12-01
Kompose down doesn't remove configmaps


rkachowski:
When running `kompose up`, many resources are created including services, pods, deployments and configmaps. When running `kompose down`, created configmaps aren't removed. This results in a fatal error when `kompose up` is run subsequently.

## Expected
`kompose down` brings down all traces of the application in the cluster. `kompose up` and `down` can be run after each other

## Actual
Process dies before all resources can be created. `kompose up` emits `FATA Error while deploying application: configmaps "<app>" already exists` 

## logs
```
$ kompose up
INFO We are going to create Kubernetes Deployments, Services and PersistentVolumeClaims for your Dockerized application. If you need different kind of resources, use the 'kompose convert' and 'kubectl create -f' commands instead.

INFO Deploying application in "default" namespace
INFO Successfully created Service: app
INFO Successfully created Service: nginx
INFO Successfully created Service: postgres
INFO Successfully created Service: redis
INFO Successfully created Deployment: app
INFO Successfully created Config Map: app-app-env
INFO Successfully created Deployment: nginx
INFO Successfully created PersistentVolumeClaim: nginx-claim0 of size 100Mi. If your cluster has dynamic storage provisioning, you don't have to do anything. Otherwise you have to create PersistentVolume to make PVC work
INFO Successfully created Deployment: postgres
INFO Successfully created Config Map: postgres-app-env
INFO Successfully created PersistentVolumeClaim: postgres of size 100Mi. If your cluster has dynamic storage provisioning, you don't have to do anything. Otherwise you have to create PersistentVolume to make PVC work
INFO Successfully created Deployment: redis

Your application has been deployed to Kubernetes. You can run 'kubectl get deployment,svc,pods,pvc' for details.

$ kompose down
INFO Deleting application in "default" namespace
INFO Successfully deleted Service: app
INFO Successfully deleted Service: nginx
INFO Successfully deleted Service: postgres
INFO Successfully deleted Service: redis
INFO Successfully deleted Deployment: app
INFO Successfully deleted Deployment: nginx
INFO Successfully deleted PersistentVolumeClaim: nginx-claim0
INFO Successfully deleted Deployment: postgres
INFO Successfully deleted PersistentVolumeClaim: postgres
INFO Successfully deleted Deployment: redis

$ kompose up
[...]

PR #900, opened 2017-12-22
Added Case for Config Map in kompose down

piyush1594:
Added Case for Config Map in kompose down. To implement this, added a label in the configMap object at the time of init.
It will delete the configMap object during `kompose down`
#883


piyush1594:
@surajnarwade @cdrage Please review !!

cdrage:
This LGTM :+1:

PR #903, opened 2017-12-23
Fix link to binary-installation method in README.md

### LINKED ISSUE #902: No proper link to Binary Installation method in README.md
Link for the **Binary (Prefered method)** method in the `README.md` opens https://github.com/kubernetes/kompose/blob/master/README.md instead of jumping to the proper section from the file.

bhavin192:
Now it links to the section of same file.
Also fixes typo in word preferred.
fixes #902

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,17 +24,17 @@ Other examples are provided in the _examples_ [directory](./examples).
 
 ## Installation
 
-We have multiple ways to install Kompose. Our prefered method is downloading the binary from the latest GitHub release.
+We have multiple ways to install Kompose. Our preferred method is downloading the binary from the latest GitHub release.
 
 Our entire list of installation methods are located in our [installation.md](/docs/installation.md) document.
 
 Installation methods:
-  - [Binary (Prefered method)](README.md)
+  - [Binary (Preferred method)](/docs/installation.md#github-release)
   - [Go](/docs/installation.md#go)
   - [CentOS](/docs/installation.md#centos)
   - [Fedora](/docs/installation.md#fedora)
   - [macOS (Homebrew)](/docs/installation.md#macos)
-  - [Windows](docs/installation.md#windows)
+  - [Windows](/docs/installation.md#windows)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,12 +24,12 @@ Other examples are provided in the _examples_ [directory](./examples).
...
 ## Installation
 
-We have multiple ways to install Kompose. Our prefered method is downloading the binary from the latest GitHub release.
+We have multiple ways to install Kompose. Our preferred method is downloading the binary from the latest GitHub release.
 
 Our entire list of installation methods are located in our [installation.md](/docs/installation.md) document.
 
 Installation methods:
-  - [Binary (Prefered method)](README.md)
+  - [Binary (Preferred method)](#binary-installation)
```
cdrage: I think this should be `/docs/installation.md#binary-installation` otherwise lgtm.

surajnarwade:
Thanks @bhavin192 for your contribution, after changes suggested by @cdrage , good to go :)

README.md
bhavin192: We don't have a section named `binary-installation` in `/docs/installation.md` (actual method is there though)
but the same installation method is present in `README.md` itself, so I thought linking to the section from `README.md` will be better solution.
should I link to `/docs/installation.md` instead?

cdrage: @bhavin192 sorry, link to: https://github.com/kubernetes/kompose/blob/master/docs/installation.md#github-release

so: `docs/installation.md#github-release`

bhavin192: @cdrage I got it now, so the section is like a index of `installation.md` but again won't it be little bit confusing for user? as we have instructions there on the `README.md` itself (for binary installation) and we are redirecting to `installation.md` for the same instructions?
Can we have list of all methods after the Binary Installation method (`README.md`) ?

cdrage:
IMO. Having it separate would be best as it shows other methods for
installation too.

cdrage:
LGTM! Thanks!

PR #907, opened 2018-01-07
Allow services to use same port with different protocols

caglar10ur:
kompose fails if compose file declares different protocols for the same port. eg;

```
...
     ports:
      - 666:666/udp
      - 666:666/tcp
...
```

This PR adds the port to the output and also makes sure that names are unique for each port/protocol pair. This is supported with LoadBalancer (https://github.com/kubernetes/kubernetes/issues/2995
) so trying to use this config with LB panics

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -337,6 +337,9 @@ func (k *Kubernetes) ConfigPorts(name string, service kobject.ServiceConfig) []a
 // ConfigServicePorts configure the container service ports.
 func (k *Kubernetes) ConfigServicePorts(name string, service kobject.ServiceConfig) []api.ServicePort {
 	servicePorts := []api.ServicePort{}
+	seenPorts := make(map[int]struct{}, len(service.Port))
+
+	var servicePort api.ServicePort
 	for _, port := range service.Port {
 		if port.HostPort == 0 {
 			port.HostPort = port.ContainerPort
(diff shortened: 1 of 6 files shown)


cdrage:
Hey @caglar10ur Seems that tests are failing.

Do you mind updating your PR with a description of what's happening here / being changed as well?

caglar10ur:
@cdrage sure thing. While at it I'll also add a fixture for this work.

caglar10ur:
@cdrage done and now tests are passing too

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -346,21 +349,29 @@ func (k *Kubernetes) ConfigServicePorts(name string, service kobject.ServiceConf
...
+		// decide the name based on whether we saw this port before
+		name := strconv.Itoa(int(port.HostPort))
+		if _, ok := seenPorts[int(port.HostPort)]; ok {
+			// https://github.com/kubernetes/kubernetes/issues/2995
+			if service.ServiceType == string(api.ServiceTypeLoadBalancer) {
+				panic("Services of type LoadBalancer cannot use both TCP and UDP for same port " + name)
+			}
+			name = fmt.Sprintf("%s-%s", name, port.Protocol)
+		}
+		name = strings.ToLower(name)
```
cdrage: I think that the output should be uppercase rather than lowercase. Only reason being is that UDP and TCP is by-default uppercase within Kubernetes. Would make more sense for the names to be: `80-UDP` or `80-UDP` by default.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -346,21 +349,29 @@ func (k *Kubernetes) ConfigServicePorts(name string, service kobject.ServiceConf
 		targetPort.IntVal = port.ContainerPort
 		targetPort.StrVal = strconv.Itoa(int(port.ContainerPort))
 
+		// decide the name based on whether we saw this port before
+		name := strconv.Itoa(int(port.HostPort))
+		if _, ok := seenPorts[int(port.HostPort)]; ok {
+			// https://github.com/kubernetes/kubernetes/issues/2995
+			if service.ServiceType == string(api.ServiceTypeLoadBalancer) {
+				panic("Services of type LoadBalancer cannot use both TCP and UDP for same port " + name)
```
cdrage: This function should really be returning an `err`, but as of right now, that's in a future refactor.

Instead of a panic, can you please use `log.Fatalf` ? Would be helpful to output which service we're using too.

```go
log.Fatalf("Service %s of type LoadBalancer cannot use TCP and UDP for the same port", name)
```

caglar10ur:
@cdrage pushed a new commit to address your comments. Thanks!

cdrage:
@caglar10ur Do you mind if you merge the two commits? Push with 1 commit? :+1:

caglar10ur:
@cdrage squashed and merged

surajnarwade:
@caglar10ur , semaphore CI is breaking :(

caglar10ur:
@surajnarwade not sure what is going on there as I didn't see anything alarming in the logs and assumed intermittent CI failure. Do you see an error that I'm not seeing or missing?

cdrage:
Sorry for that @caglar10ur, @surajnarwade is wrong. 

Our OpenShift tests have been kind of flakey lately. Ignore that test. I'll re-run it anyways!

cdrage:
This LGTM, @surajnarwade mind doing a quick review and we can merge this in? :+1:

Issue #911, opened 2018-01-19
Unable to push Docker image for service web: Unable to retrieve .docker/config.json authentication details. 


ruddra:
I got this error while running: `kompose up`
I have been using the following docker-compose.yml:

> version: '3'
> 
> services:
>   db:
>     image: mysql:5.5
>     container_name: db-container
>     ports:
>       - "3406:3306"
>     environment:
>       MYSQL_ROOT_PASSWORD: mypassword
>       MYSQL_USER: root
>       MYSQL_DATABASE: mydb
>   web:
>     build: .
>     container_name: web-container
>     command: python3 manage.py runserver 0.0.0.0:8000
>     volumes:
>       - .:/code
>     image: "ruddra/web-image"
>     ports:
>       - "8000:8000"
>     depends_on:
>       - db

I ran the following commands:
>kompose convert --provider=openshift
>kompose up --provider=openshift

I was able to push to my dockerhub repository using: 
`docker-compose push`

cdrage:
Hey @ruddra I've been trying to diagnose this issue.

I'm assuming you're using OS X?

ruddra:
I am using macOS high sierra :)

cdrage:
@ruddra That'd be why :) Docker is running in a VM and I believe they changed the configuration location. I'll have to investigate.

ruddra:
In '~/.docker/config.json'

coodix:
Same issue for me. Trying use local private repo, but even after docker login keep getting error "FATA Error while deploying application: k.Transform failed: Unable to push Docker image for service proxy: Unable to retrieve .docker/config.json authentication details. Check that 'docker login' works successfully on the command line.: Failed to read authentication from dockercfg".

~/.docker/config.json is in place

cdrage:
Hey @coodix @ruddra @kbroughton

I got time this week to get around to this issue, unfortunately, I don't have a Mac and I'm unable to run a VM (against licensing :() to diagnose it. So I need your help. 

I'd like you all (or at least one of you) to these two things:

From this issue (https://github.com/pachyderm/pachyderm/issues/2446) , let's try this:

1. Can you please output what your DOCKER_CONFIG variable is for me. If it's blank, try the command below the run Kompose again and see if it succeeds.

```sh
echo $DOCKER_CONFIG
export DOCKER_CONFIG=~/.docker/ 
```

From this issue (https://github.com/fsouza/go-dockerclient/issues/677)

[...]

PR #913, opened 2018-01-25
Push image without authentication

coodix:
For fixing issue #911

### DIFF
--- a/pkg/utils/docker/push.go
+++ b/pkg/utils/docker/push.go
@@ -59,7 +59,17 @@ func (c *Push) PushImage(fullImageName string) error {
 	// $DOCKER_CONFIG/config.json, $HOME/.docker/config.json , $HOME/.dockercfg
 	credentials, err := dockerlib.NewAuthConfigurationsFromDockerCfg()
 	if err != nil {
-		return errors.Wrap(err, "Unable to retrieve .docker/config.json authentication details. Check that 'docker login' works successfully on the command line.")
+		log.Warn(errors.Wrap(err, "Unable to retrieve .docker/config.json authentication details. Check that 'docker login' works successfully on the command line."))
+	}
+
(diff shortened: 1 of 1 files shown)


cdrage:
@coodix does this actually work on MacOS?

coodix:
@cdrage , yes, checked with local private repository which was set up it with:
`docker run -d -p 5000:5000 --restart=always --name registry registry:2`

And pushed image to localhost:5000.

pkg/utils/docker/push.go
```diff
@@ -59,7 +59,16 @@ func (c *Push) PushImage(fullImageName string) error {
 	// $DOCKER_CONFIG/config.json, $HOME/.docker/config.json , $HOME/.dockercfg
 	credentials, err := dockerlib.NewAuthConfigurationsFromDockerCfg()
 	if err != nil {
-		return errors.Wrap(err, "Unable to retrieve .docker/config.json authentication details. Check that 'docker login' works successfully on the command line.")
+		log.Warn(errors.Wrap(err, "Unable to retrieve .docker/config.json authentication details. Check that 'docker login' works successfully on the command line."))
+	}
+
+	if credentials == nil {
```
cdrage: Mind adding a comment here saying TODO: refactor as well as detail the error (add the fsouza/dockerclient-go) issue? Otherwise, this code LGTM.

coodix: The thing is that it would be nice to have this approach (try push without authentication) cause when you use local private repo it doesn't actually require login (via docker login command) there before accessing it. 

If you thing kompose should always expect available credentials in config.json I can add the comment "TODO: refactor as well as dockerclient-go starts retrieve credentials from osxkeychain (add the fsouza/dockerclient-go) issue".

What do you think?

PR #916, opened 2018-01-29
reduce binary size

oilbeater:
remove symbol tables and debug info to reduce binary size, see: https://golang.org/cmd/link/


cdrage:
Honestly, I'd rather keep this in here for now as it makes it easier for debugging the binaries :)

PR #924, opened 2018-02-06
Remove unneeded TODO comment

hangyan:
Because when we check the version of the compose file, we have already read the file's content and ensure the file exist


cdrage:
LGTM.

Issue #892, opened 2017-12-13
panic: runtime error: invalid memory address or nil pointer dereference


pkelleratwork:
version 1.6.0 (e4adfef)
macOS 10.13.2

kompose convert =

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x221af2b]

goroutine 1 [running]:
github.com/kubernetes/kompose/pkg/loader/compose.dockerComposeToKomposeMapping(0xc4205dc6c0, 0x20, 0xc42045d820, 0x1, 0x1, 0xc4200d35f0)
	/home/wikus/dropbox/dev/go/src/github.com/kubernetes/kompose/pkg/loader/compose/v3.go:323 +0x45b
github.com/kubernetes/kompose/pkg/loader/compose.parseV3(0xc4203686f0, 0x1, 0x1, 0x1, 0x1, 0x0, 0x30, 0x27e2280)
	/home/wikus/dropbox/dev/go/src/github.com/kubernetes/kompose/pkg/loader/compose/v3.go:113 +0x22b
github.com/kubernetes/kompose/pkg/loader/compose.(*Compose).LoadFile(0x370a980, 0xc4203686f0, 0x1, 0x1, 0xc4200d3080, 0x0, 0x0, 0x9, 0x7)
	/home/wikus/dropbox/dev/go/src/github.com/kubernetes/kompose/pkg/loader/compose/compose.go:186 +0x4d2
github.com/kubernetes/kompose/pkg/app.Convert(0x100000100, 0x0, 0x0, 0x0, 0x0, 0x28a1b28, 0x4, 0x0, 0x28b5b4d, 0x15, ...)
	/home/wikus/dropbox/dev/go/src/github.com/kubernetes/kompose/pkg/app/app.go:227 +0x145
github.com/kubernetes/kompose/cmd.glob..func3(0x36d8660, 0x370a980, 0x0, 0x0)
	/home/wikus/dropbox/dev/go/src/github.com/kubernetes/kompose/cmd/convert.go:95 +0x4d
github.com/kubernetes/kompose/vendor/github.com/spf13/cobra.(*Command).execute(0x36d8660, 0x370a980, 0x0, 0x0, 0x36d8660, 0x370a980)
	/home/wikus/dropbox/dev/go/src/github.com/kubernetes/kompose/vendor/github.com/spf13/cobra/command.go:704 +0x2c6
[...]

PR #926, opened 2018-02-07
Fix unset env bug for v3 compose

### LINKED ISSUE #918: "panic: runtime error: invalid memory address or nil pointer dereference" when using environment unset variables
when using a compose file containing an unset environment var, kompose crash with:
```
kompose up -f docker-compose.yml 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1621e4b]
```

example:
```
version: '3'

services:
  jenkins:
    image: jenkins
    environment:
      - DEBUG
```

Note: if the environment var is set, everything work as expected. example:
```
version: '3'

services:
  jenkins:
    image: jenkins
    environment:
      - DEBUG=1
```

hangyan:
fix #918

### DIFF
--- a/pkg/loader/compose/v3.go
+++ b/pkg/loader/compose/v3.go
@@ -319,8 +319,19 @@ func dockerComposeToKomposeMapping(composeObject *types.Config) (kobject.Kompose
 		// Gather the environment values
 		// DockerCompose uses map[string]*string while we use []string
 		// So let's convert that using this hack
+		// Note: unset env pick up the env value on host if exist
 		for name, value := range composeServiceConfig.Environment {
-			env := kobject.EnvVar{Name: name, Value: *value}
+			var env kobject.EnvVar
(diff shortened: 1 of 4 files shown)


cdrage:
Tests pass... is it possible to add a test? If not, I could quickly write one up 👍

hangyan:
Sure, I will add some unit test

?? Outlook for Android<https://aka.ms/ghei36>

________________________________

hangyan:
@cdrage I have add some tests and check the docker-compose doc and find that my previous code is not correct. Link: https://docs.docker.com/compose/compose-file/#environment. I have rewrite the logic, please help me review it

hangyan:
related to #892 . @cdrage

cdrage:
These LGTM. Thanks for the contribution!!

PR #937, opened 2018-02-09
Variable name should not be the same as imported package name

hangyan:


hangyan:
@surajnarwade @cdrage CI seems unstable..

cdrage:
@hangyan Yeah, looks like the Fedora build system went down temporarily. I've gone ahead and restarted the build.

hangyan:
@cdrage  how do i re-trigger the build?

cdrage:
@hangyan If you do a force update, it can re-trigger the build (for example, using `git commit --amend` and amending the message and updating the PR).

I've triggered a rebuild manually however..

hangyan:
@cdrage CI passed

PR #938, opened 2018-02-09
Fix custom output and charts conflict error

### LINKED ISSUE #886: Output option with convert -c fails.
Hi there,

When using `kompose convert -c` with the `-o` option, the charts are created in the desired folder but the last step fails:
```
> kompose convert -c -o chart
kompose convert -c -o chart
INFO Kubernetes file "chart/service-service.yaml" created
INFO Kubernetes file "chart/service-deployment.yaml" created
WARN open docker-compose/templates/chart/service-service.yaml: no such file or directory
WARN open docker-compose/templates/chart/service-deployment.yaml: no such file or directory
INFO chart created in "./docker-compose/"
```

The resulting `docker-compose` folder lacks the template files:
```
> tree docker-compose
docker-compose
├── Chart.yaml
├── README.md
└── templates
```

It would be nice to be able to choose the name of the outputting directory rather than it being hardcoded to `docker-compose`.

Cheers,

Jun

hangyan:
fix #886 

@cdrage I didn't find a correct way to add test for  this change. So can you help me with this?

After this change, the logic now becomes:

assume there is a yaml in /tmp/a.yaml

* -c : results in /tmp/a/
* -c -o /tmp/charts : results in /tmp/charts/ (yaml in /tmp/charts/templates)
* -o /tmp/charts: results in /tmp/charts 
*  no args: results yaml in current dir

### DIFF
--- a/pkg/transformer/kubernetes/k8sutils.go
+++ b/pkg/transformer/kubernetes/k8sutils.go
@@ -51,14 +51,11 @@ import (
 /**
  * Generate Helm Chart configuration
  */
-func generateHelm(filenames []string, outFiles []string) error {
+func generateHelm(dirName string) error {
 	type ChartDetails struct {
 		Name string
 	}
-	// Let assume all the docker-compose files are in the same directory
-	filename := filenames[0]
-	extension := filepath.Ext(filename)
-	dirName := filename[0 : len(filename)-len(extension)]
+
 	details := ChartDetails{dirName}
 	manifestDir := dirName + string(os.PathSeparator) + "templates"
 	dir, err := os.Open(dirName)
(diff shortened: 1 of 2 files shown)


cdrage:
To be honest @hangyan I don't think there are *any* tests written at the moment for charts. I'll review this soon however and test it out before we merge this in!

hangyan:
@cdrage I have add some simple tests for charts and cutom output working together, I think it's enough for now

pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -200,14 +195,25 @@ func PrintList(objects []runtime.Object, opt kobject.ConvertOptions) error {
...
 			return errors.Wrap(err, "transformer.Print failed")
 		}
 		files = append(files, printVal)
 	} else {
+		finalDirName := dirName
+		if opt.CreateChart {
+			finalDirName = dirName + string(os.PathSeparator) + "templates"
+		}
+
+		log.Infof("YAML results in : %s", finalDirName)
```
cdrage: Is this suppose to be debug information?

cdrage: If not, it may be better to indicate that this is where the chart information will be :) For example: `Chart files located at: %s", finalDirName` or something along those lines.

hangyan: @cdrage updated, remove the debug information, thanks very much!

cdrage:
Thanks so much for all these recent contributions! Tests pass and code LGTM :+1:

hangyan:
@cdrage Glad that I can help, hoping to do more contribute to this greate project

Issue #809, opened 2017-09-06
Cannot locate specified Dockerfile: Dockerfile


ylogx:
When I run `kompose up`, I get the following error:
```
FATA Error while deploying application: k.Transform failed: Unable to build Docker image for service business-api: Unable to build image. For more output, use -v or --verbose when converting.: API error (500): {"message":"Cannot locate specified Dockerfile: Dockerfile"}
```

My folder structure is: 
```
repo 
  package
    business-api
        Dockerfile
```

My docker-compose.yaml is:
```yaml
version: '2'

services:
  business-api:
    restart: always
    build: ./package/business_api
    image: index.docker.io/company/business-api
    volumes:
      - ./:/usr/src/app:ro
      - ./data:/root/data:ro
    depends_on:
      - some-other-service
```

cdrage:
Hmm, may be a bug with the pathing, you may have to manually specify by doing: `./package/business_api/Dockerfile` under build.

Try that with the `--verbose` output and let post the debug information here!

shubhamchaudhary:
Added full path in build -> dockerfile:

```
    build:
      context: ./package/business_api
      dockerfile: ./package/business_api/Dockerfile
```

Output of `sudo kompose up -v 1000`:

```
DEBU Docker Compose version: 2
DEBU Opening compose files: docker-compose.yml
DEBU [0/0] [webapp_npm]: EventType: 31
DEBU [0/0] [webapp_build]: EventType: 31
DEBU [0/0] [webapp_config]: EventType: 31
DEBU [0/0] [webapp_node_modules]: EventType: 31
DEBU [0/1] [nginx]: Adding
DEBU [0/5] [webapp]: Adding
DEBU [0/5] [business-api]: Adding
DEBU [0/6] [default]: EventType: 32
DEBU [0/6] [default]: EventType: 32
DEBU [0/6] [default]: EventType: 32
DEBU [0/6] [default]: EventType: 32
DEBU [0/6] [default]: EventType: 32
DEBU Default network found
DEBU [0/6] [default]: EventType: 32
WARN Unsupported root level volumes key - ignoring
WARN Unsupported depends_on key - ignoring
INFO Build key detected. Attempting to build and push image 'index.docker.io/example/business-api'
DEBU Compose file dir: /home/chaudhary/code/work/recs
INFO Building image 'index.docker.io/example/business-api' from directory 'business_api'
DEBU Created temporary file /tmp/kompose-image-build-853848623 for Docker image tarballing
[...]

PR #940, opened 2018-02-09
Accept nested path such as "build: ./foo/bar" in kompose up

k24:
In `kompose up`, building docker image but throws "Cannot locate specified Dockerfile" because `path.Base` removes parent directory path segments. I think that a nested path should be accepted.

For instance, in the case of #809, "./package/business_api" is modified to "business_api" by `path.Base`. However `os.Stat` verifies the existence of `service.Build` before the lines.

It's difficult to test `BuildDockerImage`, so I extracted `resolveImagePathAndName` from the func then modify it and add testing.

I have no idea why using `path.Base`. If having some reasons to make a sense, could you please write it as a comment? I will accept any reason and you can close this anytime. Thanks!

### DIFF
--- a/pkg/transformer/utils.go
+++ b/pkg/transformer/utils.go
@@ -217,12 +217,7 @@ func BuildDockerImage(service kobject.ServiceConfig, name string, relativePath s
 	}
 
 	// Get the appropriate image source and name
-	// use path.Base to get the last element of the relative build path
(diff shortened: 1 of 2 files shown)


surajnarwade:
@k24 , Please sign the CLA first

k24:
Sorry, I decline this. Thank you.

PR #954, opened 2018-03-08
Add support for compose v3.2

hangyan:
fix


cdrage:
@hangyan honestly, I don't see why not. The only differences with 3.1 and 3.2 is the long volume syntax. This LGTM.

PR #955, opened 2018-03-08
Add script for generating deb / rpm packages

cdrage:
Adds a script to generate deb / rpm packages on release day.

### DIFF
--- a/script/release.sh
+++ b/script/release.sh
@@ -27,6 +27,7 @@ usage() {
   echo " hub"
   echo " github-release"
   echo " github_changelog_generator"
+  echo " fpm"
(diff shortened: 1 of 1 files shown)


script/release.sh
```diff
@@ -56,6 +57,11 @@ requirements() {
     exit 0
   fi
 
+  if ! hash fpm 2>/dev/null; then
+    echo "ERROR: rpm required to generate deb/rpm packages. Please run 'gem install fpm"
```
hangyan: rpm --> fpm?

cdrage: yeah you're right :+1:

PR #956, opened 2018-03-09
Refactor tests script to uses vars

hangyan:
less is more ~


hangyan:
@cdrage @surajnarwade any thoughts?

cdrage:
@hangyan Woops, missed this PR. This is much cleaner! :+1: Thanks!

PR #961, opened 2018-03-20
fix pod-controller relationship mistake

jonyhy96:
fix pod-controller relationship mistake
```release-note
None
```


cdrage:
LGTM

Issue #851, opened 2017-10-13
Not creation of `-service.yaml` when `restart` is present


gdhgdhgdh:
Hi, using `1.3.0` of `kompose` and this trivial `docker-compose.yaml`:

```
frontend:
  image: registry:5000/frontend:1.0.0
  ports:
    - "12681:80"
    - "12683:443"
  environment:
    API_URL: "http://10.179.49.0:12603"
  restart: 'no'

redis:
  image: registry:5000/redis:3.0.7
  ports:
    - "12679:6379"
```

```
[root@node1 k]# kompose convert
INFO Kubernetes file "redis-service.yaml" created 
INFO Kubernetes file "frontend-pod.yaml" created  
INFO Kubernetes file "redis-deployment.yaml" created 
```

There's no service for the frontend, which I thought was a bit odd. When I remove the `restart` line completely, I get a different set of files:

```
[root@node1 k]# kompose convert
INFO Kubernetes file "frontend-service.yaml" created 
INFO Kubernetes file "redis-service.yaml" created 
INFO Kubernetes file "frontend-deployment.yaml" created 
INFO Kubernetes file "redis-deployment.yaml" created 
```

Yes, two services and two deployments is what I'd expect. Is this a bug?

surajnarwade:
@gdhgdhgdh , when you mention `restart: no`, kompose will generate pod.
for more reference: http://kompose.io/user-guide/#restart

gdhgdhgdh:
Yeah that makes sense - the question was really that I would still expect a `kind: Service` to be created even if it's pointing to a one-off pod. Is that wide of the mark?

surajnarwade:
@gdhgdhgdh , `restart: no` will give same behaviour as `job`, so it makes no sense to create service, I think

hangyan:
@surajnarwade @cdrage  I thinks the logic should goes to:

1. with ports, create service by default
2. without ports, no service. With labels, create Headless Service

This should cause no confuse any more.

cdrage:
@hangyan I agree with all those points. We should implement a label for a headless service rather than relying on "hidden" features such as `restart: no`. Labels such as specifying to use a Job instead..

surajnarwade:
+1 @hangyan

PR #979, opened 2018-04-02
Add headless service type label

hangyan:
Also remove the relation between restart and service create


hangyan:
close #851

hangyan:
@cdrage @kadel Help me review this if you guys have time.

docs/user-guide.md
```diff
@@ -297,12 +297,12 @@ The currently supported options are:
...
 | Key                  | Value                               |
 |----------------------|-------------------------------------|
-| kompose.service.type | nodeport / clusterip / loadbalancer |
+| kompose.service.type | nodeport / clusterip / loadbalancer / headless |
 | kompose.service.expose | true / hostname |
 | kompose.service.expose.tls-secret | secret name |
 | kompose.volume.size | kubernetes supported volume size |
 
-**Note**: `kompose.service.type` label should be defined with `ports` only, otherwise `kompose` will fail.
+**Note**: `kompose.service.type` label should be defined with `ports` only(except for headless service), otherwise `kompose` will fail.
```
cdrage: add spacing between only and (except)

pkg/loader/compose/utils.go
```diff
@@ -33,6 +33,9 @@ const (
 	LabelServiceExpose = "kompose.service.expose"
 	// LabelServiceExposeTLSSecret  provides the name of the TLS secret to use with the Kubernetes ingress controller
 	LabelServiceExposeTLSSecret = "kompose.service.expose.tls-secret"
+
+	// ServiceTypeHeadless...
+	ServiceTypeHeadless = "Headless"
```
cdrage: most other variables are lowercase, should we use `headless`  instead?

pkg/transformer/openshift/openshift.go
```diff
@@ -384,15 +384,17 @@ func (o *OpenShift) Transform(komposeObject kobject.KomposeObject, opt kobject.C
...
-				}
-			} else {
+		if o.PortsExist(name, service) {
+			svc := o.CreateService(name, service, objects)
+			objects = append(objects, svc)
+
+			if service.ExposeService != "" {
+				objects = append(objects, o.initRoute(name, service, svc.Spec.Ports[0].Port))
+			}
+		} else {
```
cdrage: you can use an `else if` here :)

pkg/loader/compose/utils.go
hangyan: This is a special case. Since `api.ServiceTypeNodePort/api.ServiceTypeClusterIP...` use `NodePort` `LoadBalancer`... So i think my be follow there rules may be better

hangyan: This is a special case. Since kubernetes api does not have a "Headless' type, it has `api.ServiceTypeClusterIP=ClusterIP`.... This is var is only for internal use, others are custom labels that will show in the compose file

pkg/transformer/openshift/openshift.go
hangyan: Updated. Thanks!

docs/user-guide.md
hangyan: Updated!, thanks!

PR #984, opened 2018-04-15
Typo fix in glide.yaml

AdamDang:
Line 80: 
"it this case it is OK"->"in this case it is OK"


hangyan:
There is no need to create sperate PRs to fix the typo. Please merge the commits into one PR and do not do this again! Will close this PR.

PR #995, opened 2018-04-27
Correct the proper nouns to be in capital letter

AdamDang:
The proper nouns like: Kubernetes, OpenShift, Transformer should be in capital letters, while there some places in the doc use lowercases.


hangyan:
@AdamDang  It's better to keep same, but we should also consider the context. For example, `transformer` means `transformer.go` file here. Thanks for the contribution anyway, but i'm gonna close this PR

PR #997, opened 2018-05-02
Remove go1.6 support in ci

hangyan:
From the error logs of the build, the root cause is that golint have abandon the support for 1.6, and because golang already have version1.9/1.10, I think we can do that too.

https://travis-ci.org/kubernetes/kompose/jobs/373755156
@cdrage


cdrage:
This LGTM. Thanks @hangyan !

PR #999, opened 2018-05-04
Support compose global deploy mode to daemonset

xianlubird:
Support compose global deploy mode to daemonset

@cdrage @hangyan Pls help to review if you are free. Thanks.

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -626,8 +626,14 @@ func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.Servic
 
 	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
 	if service.DeployMode == "global" {
-		log.Warning("Global mode not yet supported, containers will only be replicated once throughout the cluster. DaemonSet support will be added in the future.")
-		replica = 1
+		//default use daemonset
+		if opt.Controller == "" {
+			opt.CreateD = false
+			opt.CreateDS = true
+		} else if opt.Controller != "daemonset" {
+			log.Warnf("Global deploy mode service is best converted to daemonset, now it convert to %s", opt.Controller)
+		}
+
 	}
 	if opt.CreateD || opt.Controller == "deployment" {
 		objects = append(objects, k.InitD(name, service, replica))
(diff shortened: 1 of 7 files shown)


hangyan:
What if the use use `--deployment` args at the same time?

xianlubird:
@hangyan Nice catch ! Already change code, thanks.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -626,8 +626,11 @@ func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.Servic
 
 	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
 	if service.DeployMode == "global" {
-		log.Warning("Global mode not yet supported, containers will only be replicated once throughout the cluster. DaemonSet support will be added in the future.")
-		replica = 1
+		opt.CreateD = false
+		opt.CreateDS = true
+		if opt.Controller != "" && opt.Controller != "daemonset" {
+			log.Fatal("Global deploy mode can only convert to daemonset")
```
hangyan: I think fatal log is not a good idea. May be we should try to let user proceed anyway, wether use deployment( ignore deploy mode) or use ds(ignore flags). I think there should be some existing example to consult in the code.

xianlubird: Thank you for review code on weekend. Already change code to use `daemonset` controller as default when global mode get in.   @hangyan

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -626,8 +626,9 @@ func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.Servic
 
 	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
 	if service.DeployMode == "global" {
-		log.Warning("Global mode not yet supported, containers will only be replicated once throughout the cluster. DaemonSet support will be added in the future.")
-		replica = 1
+		opt.CreateD = false
+		opt.CreateDS = true
+		opt.Controller = "daemonset"
```
hangyan: At warn message is needed if the user specific a `--deployment` flag.Also in my opinion, it's better to use deployment if the user provide a `--deployment` flag. Because if we ignore it,  we can only create a daemonset if `deployMode==global`, there is no way to overwrite it. But on the other side, we can create daemonset(by default) and deployment(user provide deployment flag), I think this will give the user more choices.

xianlubird: Sure, agree this opinion.  Change code to that if user use `--deployment` flag, it will convert to `Deployment` and default to use `StatefulSet`.   Also add test case to this situation. @hangyan

xianlubird:
@cdrage @hangyan Any update on this ?

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -625,9 +625,9 @@ func (k *Kubernetes) CreateKubernetesObjects(name string, service kobject.Servic
 	}
 
 	// Check to see if Docker Compose v3 Deploy.Mode has been set to "global"
-	if service.DeployMode == "global" {
-		log.Warning("Global mode not yet supported, containers will only be replicated once throughout the cluster. DaemonSet support will be added in the future.")
-		replica = 1
+	if service.DeployMode == "global" && opt.Controller == "" {
+		opt.CreateD = false
+		opt.CreateDS = true
```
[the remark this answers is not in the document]
xianlubird: @hangyan Done. Already add warn message

PR #1007, opened 2018-05-15
Add support for file based secret

hangyan:
related to #296 

Notes:
1. External secret is not supported
2. may have conflict with #994 
3. import a new package cast(https://github.com/spf13/cast)


cdrage:
@hangyan Please separate the commits between code change and vendor update so we can review!

hangyan:
@cdrage Done. Thanks!

pkg/loader/compose/v3.go
```diff
@@ -221,12 +222,39 @@ func parseHealthCheck(composeHealthCheck types.HealthCheckConfig) (kobject.Healt
...
+		}
+	}
+	if len(result) == 0 {
+		return nil
+	} else {
+		return result
+	}
+
+}*/
+
```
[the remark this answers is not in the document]
hangyan: @cdrage This is ready for review. But i have started too long ago and forgot to remove some useless code before this review. Sorry about that! Will be more carefully about this in the future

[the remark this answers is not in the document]
cdrage: Since this code is no longer needed, please remove the commented out code :+1: @hangyan

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -285,6 +286,37 @@ func (k *Kubernetes) initIngress(name string, service kobject.ServiceConfig, por
...
 }
 
+// CreateSecrets create secrets
+func (k *Kubernetes) CreateSecrets(komposeObject kobject.KomposeObject) ([]*api.Secret, error) {
+	var objects []*api.Secret
+	for name, config := range komposeObject.Secrets {
+		if config.File != "" {
+			data, err := GetSecretDataFromFile(config.File, k.Opt)
+			if err != nil {
+				log.Fatal("unable to read secret from file:", config.File)
```
cdrage: add spacing after "file:"

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -285,6 +286,37 @@ func (k *Kubernetes) initIngress(name string, service kobject.ServiceConfig, por
...
+				ObjectMeta: api.ObjectMeta{
+					Name:   name,
+					Labels: transformer.ConfigLabels(name),
+				},
+				Type: api.SecretTypeOpaque,
+				Data: map[string][]byte{name: data},
+			}
+			objects = append(objects, secret)
+		} else {
+			log.Warnf("External secrets %s is not supported by now - ignoring", name)
```
cdrage: small grammar mistake, should be "is not currently supported"

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -429,6 +461,56 @@ func (k *Kubernetes) ConfigTmpfs(name string, service kobject.ServiceConfig) ([]
...
+
+			volMount := api.VolumeMount{
+				Name:      vol.Name,
+				MountPath: "/run/secrets",
+			}
+			volumeMounts = append(volumeMounts, volMount)
+		}
+	}
+	return volumeMounts, volumes
+}
```
cdrage: Are you able to comment on some of the code above as to what's happening?

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -697,6 +783,16 @@ func (k *Kubernetes) Transform(komposeObject kobject.KomposeObject, opt kobject.
 	// this will hold all the converted data
 	var allobjects []runtime.Object
 
+	if komposeObject.Secrets != nil {
+		secrets, err := k.CreateSecrets(komposeObject)
+		if err != nil {
+			return nil, errors.Wrapf(err, "create secrets error")
```
cdrage: Maybe add something different such as "Unable to create secrets" or something more descriptive? Not too sure what type of error message to add.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -449,6 +531,10 @@ func (k *Kubernetes) ConfigVolumes(name string, service kobject.ServiceConfig) (
 		useHostPath = true
 	}
 
+	secretsVolumeMounts, secretsVolumes := k.ConfigSecretVolumes(name, service)
```
cdrage: Add comment above here explaining that we are adding secrets to volumes

cdrage:
Progress @hangyan ?

hangyan:
@cdrage Sorry, been very busy these days on my work. Working on this now.

hangyan:
@cdrage Updated

hangyan:
conflict resolved

jvitor83:
@cdrage is this good? will be merged?
I need that feature.

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -516,6 +548,60 @@ func (k *Kubernetes) ConfigTmpfs(name string, service kobject.ServiceConfig) ([]
...
+
+			vol := api.Volume{
+				Name:         secretConfig.Source,
+				VolumeSource: volSource,
+			}
+			volumes = append(volumes, vol)
+
+			volMount := api.VolumeMount{
+				Name:      vol.Name,
+				MountPath: "/run/secrets",
```
jvitor83: @hangyan 
This shouldn't be `MountPath: "/run/secrets/" + vol.Name,` ???

Because without the MountPath targeting to the secret, it gives a error:
```
RunContainerError: failed to start container "f97d9e9354548939d15acae70758347538426fbb05c80f2a5cd316dc67e7bf58": 

Error response from daemon: oci runtime error: container_linux.go:247: starting container process caused "process_linux.go:359: 

container init caused \"rootfs_linux.go:54: 
mounting \\\"/opt/rke/var/lib/kubelet/pods/516c135b-bce8-11e8-8714-005056ae708d/volumes/kubernetes.io~secret/default-token-khnzc\\\" 
to rootfs \\\"/var/lib/docker/overlay/60fb5d9a4007bf8e1889898df737b275cb77431351b9168ab4d629fbcdf065b5/merged\\\" 
at \\\"/var/lib/docker/overlay/60fb5d9a4007bf8e1889898df737b275cb77431351b9168ab4d629fbcdf065b5/merged/run/secrets/kubernetes.io/serviceaccount\\\" 
caused \\\"mkdir /var/lib/docker/overlay/60fb5d9a4007bf8e1889898df737b275cb77431351b9168ab4d629fbcdf065b5/merged/run/secrets/kubernetes.io: read-only file system\\\"\""
```
Just like in this issue: https://github.com/kubernetes/kubernetes/issues/65835

And concating the secret name, it work!

jvitor83: With some research i discover that:

## kubernetes
```bash
kubectl create secret generic my_secret --from-literal=my_secret_key_1=my_secret_value_1 --from-literal=my_secret_key_2=my_secret_value_2
```
it create a file `/run/secrets/my_secret/my_secret_key_1` (with the content `my_secret_value_1`)

## docker-compose/swarm

- with short syntax:
```yaml
...
    secrets:
      - my_secret
secrets:
  my_secret:
    file: my_secret.txt
```
it create a file `/run/secrets/my_secret`
> **don't allow** to be compatible with kubernetes

- with long syntax:
```yaml
...
    secrets:
      - source: some_secret
        target: my_secret/my_secret_key
secrets:
  some_secret:
    file: my_secret.txt
```
it create a file `/run/secrets/my_secret/my_secret_key`
> **allow** to be compatible with kubernetes

---

I believe that this PR:
- **shouldn't allow to use short-syntax**, or if allow, to at least concat the secret name as directory (as my previous comment suggest to avoid the error).
- **only allow long-syntax if has at least inside one directory** to be compatible with kubernetes

jvitor83:
@hangyan , i create a PR to your secret-support branch (PR) which:
- Fix the issue reported at https://github.com/kubernetes/kompose/pull/1007#discussion_r219224280
- Added support to long-syntax with absolute path

hangyan:
@jvitor83 Great. Thanks very much. I will review this again ASAP. But since there are not so many active maintainers for this project, this PR may take a long time to be merged in. I'm afraid of you will have to build the binary yourself based on this branch.

jvitor83:
> I'm afraid of you will have to build the binary yourself based on this branch.

Already did it!
Thanks for the feedback.

cdrage:
@hangyan if you want @jvitor83 you can always do a branch of this PR, push your changes and open a new PR. I don't think @hangyan would mind if you continue on his awesome work :+1:

pkg/transformer/kubernetes/kubernetes.go
hangyan: @jvitor83 Sorry about the late response. I have re-checked the document of kubernetes an docker. I have different opinion on this problem. kompose is meant to translate compose to kubernetes as precision as possible. Because docker mount the secret to `/run/secrets/<secret-name>`, so this path should be the same in kubernetes, not `/run/secrets/<secret-name>/<key>`. I believe the right choice is to use subPath ( you can see a example in the issues link you provide)

jvitor83: I get it that _"kompose is meant to translate compose to kubernetes as precision as possible"_, but it didn't work with the k8s generated!
The generated mapping gives error (mentioned before) and the pod didn't start!

By my understanding, the generated files should work in first place to then keep the translation as precise as possible.

I guess we should remove the short syntax once it is incompatible and don't allow to get the file at the same path.

hangyan: The issue link you provide has a solution in the end:subPath. I think it's a better solution, i haven't test it out, but I think it should work

jvitor83: @hangyan 
I am not understanding what you mean.

The thing is that this PR is generating:
```yaml
        volumeMounts:
        - name: secret-name
          mountPath: "/run/secrets"
```
which gives the error reported.
**The issue is solved not because it put the `subPath`, but because the directory was added to the mountPath. (I think the issue author have been mistaken)**
If the kompose generate the k8s without the mountPath with a directory (with secret using short-syntax), it will give error on start. _I have tested it with openjdk, dotnet and others. All gives the same error._

What you are suggesting to do? 
_If is to: "modify the code to add the subPath", it will continue to give the error._

I think we have to **at least put some warning** saying that secrets with short-syntax can gives error.

PR #1016, opened 2018-05-24
Add security contacts

cdrage:
As per
https://github.com/kubernetes/community/blob/master/committee-steering/governance/sig-governance-template-short.md
a SECURITY_CONTACTS file has been created.


sebgoa:
fine with me, even though I have been very active lately.

PR #1026, opened 2018-06-03
Fix all typos in changelog

hangyan:


cdrage:
@hangyan this is actually automatically generated by our release script (https://github.com/github-changelog-generator/github-changelog-generator) so it'll still be overwritten each time as it uses previous / past commits. The only way to change this would be to forcefully go through each commit message and update it so that there is no typo (which we don't want to do)

hangyan:
@cdrage Got it. ( At first I just want to fix it to prevent other people keeping create typo PR for this file)

PR #1039, opened 2018-06-12
Support to label kompose.service.image-pull-policy

jvitor83:
This PR gives the label: `kompose.service.image-pull-policy`

resolve https://github.com/kubernetes/kompose/issues/1036


jvitor83:
Closing my PR since https://github.com/kubernetes/kompose/pull/1040 appear to do the same (only with different label).

Issue #1062, opened 2018-08-11
dots get propagated into invalid names


rcmorano:
Aloha,

I've been converting some compose files to Helm charts (concretely [DAppNode's project](https://github.com/dappnode/DNP_VPN/blob/master/docker-compose-vpn.yml) ones) and ran into the same problem exposed in #420.
In fact it's [the last comment from @xiaoping378 in this issue.](https://github.com/kubernetes/kompose/issues/420#issuecomment-319583013)

I've created #1063  PR following the guidelines of previous PRs #509 and #569 which fixed the problem and added the needed warning bits to documentation.
I Hope you find it's suitable to merge or as base for future work on the issue :)

Best!

PR #1063, opened 2018-08-11
Propagate dots into valid service names

rcmorano:
Fix approach for #1062; a similar bug than #420.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -429,7 +429,7 @@ services:
 
 If the Docker Compose file has a volume specified for a service, the Deployment (Kubernetes) or DeploymentConfig (OpenShift) strategy is changed to "Recreate" instead of "RollingUpdate" (default). This is done to avoid multiple instances of a service from accessing a volume at the same time.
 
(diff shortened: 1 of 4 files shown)


rcmorano:
CLA signed!

docs/user-guide.md
```diff
@@ -429,7 +429,7 @@ services:
 
 If the Docker Compose file has a volume specified for a service, the Deployment (Kubernetes) or DeploymentConfig (OpenShift) strategy is changed to "Recreate" instead of "RollingUpdate" (default). This is done to avoid multiple instances of a service from accessing a volume at the same time.
 
-If the Docker Compose file has service name with `_` in it (eg.`web_service`), then it will be replaced by `-` and the service name will be renamed accordingly (eg.`web-service`). Kompose does this because "Kubernetes" doesn't allow `_` in object name.
+If the Docker Compose file has service name with `_` or `.` in it (eg.`web_service` or `web.service`), then it will be replaced by `-` and the service name will be renamed accordingly (eg.`web-service`). Kompose does this because "Kubernetes" doesn't allow `_` in object name nor metadata name (rises "DNS-1035 label must consist of lower case alphanumeric characters").
```
hangyan: Isn't object name and metadata name the same thing?

hangyan: Others LGTM

rcmorano: Right! 

I made the distinction because the original documentation refered to a problem with the "spec" part of the kube file, but at the end, metadata is just another Kubernetes object so I guess it can be simplified to "object name" :)

hangyan:
@rcmorano Thanks

PR #1073, opened 2018-10-21
Bump Go versions

HaraldNordgren:
Appears the the fetching of `lint` was recently broken and this appears to be a fix: https://github.com/golang/lint/issues/415

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,14 +9,14 @@ go_import_path: github.com/kubernetes/kompose
 
 matrix:
   include:
-    - go: 1.7
+    - go: "1.9"
       env: 
         # Only build docs once
         - BUILD_DOCS=yes
         # Test cross-compile as well
         - CROSS_COMPILE=yes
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,14 +9,14 @@ go_import_path: github.com/kubernetes/kompose
 
 matrix:
   include:
-    - go: 1.7
+    - go: "1.9"
```
cdrage: mind removing the quotes? otherwise this LGTM!

HaraldNordgren: Travis requires (or used to require, this may be fixed by now) minor versions with two digits (10, 11, etc.) to have quotes, or use the format 1.10.x and 1.11.x.

Otherwise it incorrectly parses them as

    1.10 -> 1.1
    1.11 -> 1.1

when starting the builds.

cdrage: ah that makes sense. no worries then!

PR #1074, opened 2018-10-28
Bump Go versions and use '.x' to always get latest patch versions

HaraldNordgren:


cdrage:
Honestly I'd rather keep the hard-coded versions only because if you include `.x` it *may* update during someone's PR and cause them to get a false-positive.

hangyan:
I don't have a strong opnion on this, both looks good to me.

PR #1075, opened 2018-11-05
1.17.0 Release

cdrage:


hangyan:
@cdrage Sorry I have been really busy on my own work these days, barely have time to keep working on kompose. After 2019.1.1, I will return to keep working on this project.

cdrage:
@hangyan No worries man! To be honest, it looks like Docker is moving away from Docker Compose, so I wouldn't worry much. The work that's been going into Kompose is still going great and covering all the new features they're adding to Compose.

PR #1076, opened 2018-11-06
add annotations to ingress

### LINKED ISSUE #1071: Labels not converted to annotations in ingress
If you have something like:

```yaml
version: '2'
services:
    web:
        labels:
            kompose.service.expose: dev.example.com
            kompose.service.expose.tls-secret: tls-secret
            kompose.service.type: clusterip

            kubernetes.io/ingress.class: nginx
            certmanager.k8s.io/cluster-issuer: letsencrypt-prod
            nginx.ingress.kubernetes.io/rewrite-target: /
        ports:
            - '80:80'
```

the annotations related to the ingress controller are not present in the `Ingress` object (they're there in `Service` and `Deployment`).

treksler:
this closes #1071

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -332,6 +332,7 @@ func (k *Kubernetes) initIngress(name string, service kobject.ServiceConfig, por
 		ObjectMeta: api.ObjectMeta{
 			Name:   name,
 			Labels: transformer.ConfigLabels(name),
+			Annotations:  transformer.ConfigAnnotations(service),
 		},
 		Spec: extensions.IngressSpec{
 			Rules: []extensions.IngressRule{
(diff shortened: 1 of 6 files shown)


hangyan:
@treksler Would you mind sign the CLA first?

treksler:
I signed it

cdrage:
This LGTM

cdrage:
Hi @treksler can you rebase so we can make sure it passes tests?

Many thanks!

mikegerber:
This would be useful for me, as it should fix issue #1071.

stevefan1999-personal:
Reason for reviving: Some specific use case. This is really useful for people like me who combines traefik and kompose to effortlessly convert legacy docker compose services with the use of external-dns, which it unfortunately looks for annotations in ingress kind only, not on deployment, which is the current behavior to convert docker compose labels. 

So I either had to re-patch the ingress router after installing the helm chart created by kompose, or I will have to dump the chart and then patch it inline. Both of them are very tedious.

hangyan:
> Reason for reviving: Some specific use case. This is really useful for people like me who combines traefik and kompose to effortlessly convert legacy docker compose services with the use of external-dns, which it unfortunately looks for annotations in ingress kind only, not on deployment, which is the current behavior to convert docker compose labels.
> 
> So I either had to re-patch the ingress router after installing the helm chart created by kompose, or I will have to dump the chart and then patch it inline. Both of them are very tedious.

Sorry for you to wait so long, i will look into it

PR #1092, opened 2019-01-03
Support multiple hostnames on expose label in service for Kubernetes

huikaihoo:
Support multiple hostnames [separated by semicolon] on expose label (kompose.service.expose) in service for Kubernetes.

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -324,6 +324,8 @@ func (k *Kubernetes) InitDS(name string, service kobject.ServiceConfig) *extensi
 
 func (k *Kubernetes) initIngress(name string, service kobject.ServiceConfig, port int32) *extensions.Ingress {
 
+	hosts := strings.Split(service.ExposeService, ";")
+
 	ingress := &extensions.Ingress{
 		TypeMeta: unversioned.TypeMeta{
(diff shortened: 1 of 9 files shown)


cdrage:
I believe we should use comma (,) rather than semicolon (;).

I rarely see semicolons be used.

Could you also test against using, for example, spaces in between? `example.com, examplefoobar.com`, etc?

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -324,6 +324,8 @@ func (k *Kubernetes) InitDS(name string, service kobject.ServiceConfig) *extensi
 
 func (k *Kubernetes) initIngress(name string, service kobject.ServiceConfig, port int32) *extensions.Ingress {
 
+	hosts := strings.Split(service.ExposeService, ";")
```
cdrage: We should also use some form of checking (making sure spaces are removed, that someone doesn't add: `,,` for example..

cdrage:
Oh! Could you add some documentation too for when this is merged? We should indicate to the users that multiple values are allowed.

huikaihoo:
Updated. Please take a look at it.

cdrage:
Awesome on the tests! :+1: and the code LGTM. 

Thanks for the contribution!

PR #1109, opened 2019-03-02
fix: hostpath volume supoort for docker-compose v3 file

### LINKED ISSUE #1089: HostPath volumes incorrectly concatenate paths
When using `--volumes hostPath`, the path of the volume (at least on macOS) incorrectly concatenates the full path of the docker-compose.yml file to the full path of the host mount:

Current:
```yaml
volumes:
      - hostPath:
          path: /Users/foobar/Documents/Projects/x/Users/foobar/Documents/Projects/x/runtime-media/mysql
        name: mysql-hostpath0
```

Expected:
```yaml
volumes:
      - hostPath:
          path: /Users/foobar/Documents/Projects/x/runtime-media/mysql
        name: mysql-hostpath0
```

See: https://github.com/kubernetes/kompose/pull/957/

ZedYeung:
fix #1089 
The problem is that in v3, we use the "long syntax" for volumes in terms of parsing.
So the path is already an absolute path. We don't need to join it with dir.
Therefore, the idea is to simply get the docker-compose file version and use switch case.


ZedYeung:
Already signed.

PR #1114, opened 2019-03-07
Calculate the absolute path directly instead of join (Resolves #1089)

### LINKED ISSUE #1089: HostPath volumes incorrectly concatenate paths

slowr:
HostPath volume path currently is calculated wrong by concatenating the compose's absolute path with the one defined in the file.

Instead I calculate the absolute path directly and add that instead of join.

Resolves #1089


slowr:
@hangyan any update on this PR's status?

PR #1123, opened 2019-04-03
Add node references

cdrage:
Removes old Kubernetes references to master / slave and converts them to
master / node.


hangyan:
Is this really necessary...

cdrage:
@hangyan Yes, we were using old references, see: https://kubernetes.io/docs/concepts/overview/components/ Kubernetes now uses Master Node / Node.

Once tests pass I'll merge this in :+1:

PR #1141, opened 2019-06-19
Fix container_name set error

hangyan:


hangyan:
fix #1135

cdrage:
Code looks good.

/lgtm

Issue #1148, opened 2019-06-25
Travis ci seems broken


hangyan:
https://travis-ci.org/kubernetes/kompose/jobs/548728449

it ignores the test errors in 'test-cmd'

hangyan:
@cdrage

PR #1149, opened 2019-06-25
Fix test-cmd errors

hangyan:
* fix stdin error
 * fix container name error
 * fix config load error


hangyan:
fix #1148

hangyan:
There are three porbelems:

1. #1141 introduces some test errors 
2. service config's file path become abs path now(does't know when and why,) breaks some tests
3. stdin support, the code and my local test is ok now, does't know it failed on traivs

cdrage:
Woops.

Thanks for this fix!

/lgtm

hangyan:
Does the lgtm label not working?

Issue #1150, opened 2019-06-26
Label `kompose.service.type: loadbalancer` does not convert to type LoadBalancer


cjancsar:
Version Kompose `1.18.0 (06a2e56)`

My relevant docker_compose service looks like:

```
  gateway:
    image: "${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/${ECR_GATEWAY_REPO}:${TAG}"
    build: "${GATEWAY_MONOREPO_DIR}"
    ports:
      - "33000:4200"
    deploy:
      mode: replicated
      replicas: 1
      endpoint_mode: vip
    labels:
      kompose.service.type: loadbalancer
```

I have also seen examples where it is `LoadBalancer` vs `loadbalancer` (the docks show `loadbalancer`).

This is converted to a NodePort type:

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    kompose.cmd: C:\ProgramData\chocolatey\lib\kubernetes-kompose\tools\kompose.exe
      convert
    kompose.service.type: loadbalancer
    kompose.version: 1.18.0 (06a2e56)
  creationTimestamp: null
  labels:
    io.kompose.service: gateway
  name: gateway
spec:
  ports:
  - name: "33000"
    port: 33000
    targetPort: 4200
  selector:
    io.kompose.service: gateway
  type: NodePort
status:
  loadBalancer: {}
```

Is this functionality broken again?

cjancsar:
The cause of this is when `endpoint_mode=VIP` is set on a service then it seems to break the conversion or forces the service type to `NodePort`. This should probably be added to notes of `endpoint_mode` on https://github.com/kubernetes/kompose/blob/master/docs/conversion.md

hangyan:
@cjancsar Yes, you were right! Thank you for pointing it out. I will create a PR to add this note

PR #1151, opened 2019-06-26
Add node about endpoint_mode in doc

hangyan:


hangyan:
fix #1150

/assign cdrage

PR #1161, opened 2019-08-07
Threat network as namespace

jvitor83:
This PR should resolve https://github.com/kubernetes/kompose/issues/1066 by converting the network as namespace.
- All objects generated for the specific service will have the namespace set as the network name
- When have `external: true` it not create the namespace yaml

PS: I cound't test for openshift since i have no knowledge of it, but i make the same change since the logic is the same. May need some help for test for os!


jvitor83:
I made a append to the git commit to remove the `fix`, but i see the @k8s-ci-robot didn't remove the merge block. Will wait for the mantainers to check this.

nikhita:
@jvitor83 :wave: Looks like the commit message in https://github.com/kubernetes/kompose/commit/9bc491226515a22862165ff8ddae63c9c2c9203c still contains the word "fix". Can you `git amend` again? Thanks!

jvitor83:
I will remake this PR. Closing for now.

PR #1163, opened 2019-08-14
Treat network as namespace

jvitor83:
This PR should resolve https://github.com/kubernetes/kompose/issues/1066 by converting the network as namespace.
- All objects generated for the specific service will have the namespace set as the network name
- When have `external: true` it not create the namespace yaml

_PS: I cound't test for openshift since i have no knowledge of it, but i make the same change since the logic is the same. May need some help for test for openshift!_

### DIFF
--- a/pkg/kobject/kobject.go
+++ b/pkg/kobject/kobject.go
@@ -75,6 +75,7 @@ type ServiceConfig struct {
 	Args             []string            `compose:"args"`
 	VolList          []string            `compose:"volumes"`
 	Network          []string            `compose:"network"`
+	Namespace        Namespace           `compose:""`
 	Labels           map[string]string   `compose:"labels"`
 	Annotations      map[string]string   `compose:""`
 	CPUSet           string              `compose:"cpuset"`
@@ -127,6 +128,12 @@ type HealthCheck struct {
 	Retries     int32
 	StartPeriod int32
 	Disable     bool
+}
+
+// Namespace holds configuration for a service namespace
+type Namespace struct {
+	Name   string
+	Create bool
 }
 
 // EnvVar holds the environment variable struct of a container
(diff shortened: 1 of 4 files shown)


hangyan:
The code looks good. But there are still a few things to be done.

1. the openshift part, it's very similar to kubernetes
2. Update doc 
3. Update tests files

jvitor83:
Thanks @hangyan 
I will try to resolve this ASAP.

About the item 1 - The openshift part is already done (similar to kubernetes)! _The thing is about the test to see if is working, because i don't have much knowledge about openshift. But in theory the change i made to the openshift part "should work" already!_

cdrage:
Hey @hangyan and @jvitor83 

So sorry about the delays. I'm going to start doing at least some maintenance / contributing back to Kompose.

Don't worry about the OpenShift part. That part wont be necessary. But do you mind opening up a separate issue with regards to implementing it for OpenShift?

Is this PR good to go? A quick review and the code looks straight forward to me..

Could you also perhaps update the table of compatible keys, and have you written a test for Kompose yet?

hangyan:
conflict with #1189. This is a tough choice, since they are not directly mapped. I think it's better to provide user a choice, controlled by a flag, then the user can choose to convert the network part to namespace or networkpolicy.

jvitor83:
> conflict with #1189. This is a tough choice, since they are not directly mapped. I think it's better to provide user a choice, controlled by a flag, then the user can choose to convert the network part to namespace or networkpolicy.

@hangyan i already believe the https://github.com/kubernetes/kompose/issues/1188 is a better approach than this. I am thinking in close this PR if the result of https://github.com/kubernetes/kompose/issues/1188 could be achieved.
Did you think the same? I can close this?

hangyan:
@jvitor83 Yes, of course. I believe the namespace approch still make sense in some situations, it's a good idea. Thanks for you contribution.

PR #1166, opened 2019-08-21
Copy the service labels also to Deployment Pod spec annotations

alanjds:
At GetNinjas we are trying Kompose.

Having our Kubernetes configured with kube2iam, the Pods needs to have the IAM role annotated on its metadata, as said on its readme:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: aws-cli
  labels:
    name: aws-cli
  annotations:
    iam.amazonaws.com/role: role-arn
    iam.amazonaws.com/external-id: external-id
spec:
  containers:
  - image: fstab/aws-cli
    command:
      - "/home/aws/aws/env/bin/aws"
      - "s3"
      - "ls"
      - "some-bucket"
    name: aws-cli
```
The way for this to happen is to include the labels from docker-compose service into the deployment template for pods, as this PR does.

(cc: @OtavioHenrique @adelsjnr)


hangyan:
@alanjds This modification will break the tests, because it will add many auto-generated annotations and others to the pod template, you can check the travis test result.

I think a better approach would be write a specific function for this situation: just add the iam annotations if they exist, thus you do not need to update all the tests.

alanjds:
> This modification will break the tests

Yes. For some reason, running the tests locally had not showed this. I still need to become a better Golang dev 😉. 

Will fix and push again. Thanks.

> (...)  it will add many auto-generated annotations and others to the pod template. (...) I think a better approach would be write a specific function for this situation

But will the Kompose need specific rules coupled with 3rd-party projects?

In fact, by the docs I would expect every label to propagate to the Pods. Got surprised when it had not happened. This is just the 1st need of annotations we got and I would not think it will be better to do specific code for external plugins.

hangyan:
Yes, on some level i agree with you too. But doing this you need to a lot of fix to the tests file...

alanjds:
No problem. I can do some money job :)

Il gio 22 ago 2019, 23:00 Hang Yan <notifications@github.com> ha scritto:

alanjds:
@hangyan do you know how to use `make gen-cmd`? I already had read the development.md but am still confused

hangyan:
`make test-cmd` should be enough.

hangyan:
You can run the test command locally to see the difference

hangyan:
@alanjds The tests is kind of mess for now, there are still some false result on the tests but the script ignore it. You can check the details on the traivs page. But since the code change is simple , you can either dig into the tests error to see what's wrong or i can just merge this in and fix that later

alanjds:
> @alanjds The tests is kind of mess for now, there are still some false result on the tests but the script ignore it. You can check the details on the traivs page. But since the code change is simple , you can either dig into the tests error to see what's wrong or i can just merge this in and fix that later

No problem. I can send another PR after this one.

PR #1198, opened 2019-12-02
fix tests related to secrets

ChanderG:
1. fix the path used to read secret data
without this fix, the path to the compose file is added to the already
absolute path, hence readFile was unable to find the secrets file

2. fix the expected json format
mountPath includes the entire path including the name of the secret


hangyan:
Thanks @ChanderG

Issue #1131, opened 2019-05-21
Error when targeting a different (than 8080) API port


greyarch:
I am useing k3s and by default the kubernetes API listens on port `6443`. When trying to do 
```kompose up```
in this situation I get
```Error while deploying application: Get http://localhost:8080/api: dial tcp [::1]:8080: connect: connection refused```
Is there a way to target a different (than 8080) port?

**Comment by greyarch** [comment]
A simple solution is to start a proxy to the API that listens on 8080.
```
kubectl proxy --port=8080
```

**Comment by cdrage** [comment]
Yeah, this is critical and needs to be fixed. Thank you @greyarch

**Comment by loganknecht** [comment]
Hello!

@cdrage I am not sure if this is the right place to ask this. I am incredibly new to both `Kubernetes` and `Kompose`

I have a `docker-compose.yml` file that does work locally without issue and I was under the impression I would be able to use `kompose up` on it to spin up my deployment on my `minikube` locally.

However when I run it I get this output
```
╭─lknecht@愛と支えボックス ~/Repositories/api_webserver/api_webserver ‹feature/create_minikube_deployment-dev*›
╰─$ kompose up --file continuous_integration/local_api_docker_compose.yml                                                                                                                                                               1 ↵
INFO Build key detected. Attempting to build and push image 'lab/api'
INFO Building image 'lab/api' from directory 'api_webserver'
INFO Image 'lab/api' from directory 'api_webserver' built successfully
INFO Pushing image 'lab/api:latest' to registry 'docker.io'
INFO Attempting authentication credentials 'https://index.docker.io/v1/
INFO Successfully pushed image 'lab/api:latest' to registry 'docker.io'
INFO Build key detected. Attempting to build and push image 'lab/api'
INFO Building image 'lab/api' from directory 'api_webserver'
INFO Image 'lab/api' from directory 'api_webserver' built successfully
INFO Pushing image 'lab/api:latest' to registry 'docker.io'
INFO Attempting authentication credentials 'https://index.docker.io/v1/
INFO Successfully pushed image 'lab/api:latest' to registry 'docker.io'
INFO We are going to create Kubernetes Deployments, Services and PersistentVolumeClaims for your Dockerized application. If you need different kind of resources, use the 'kompose convert' and 'kubectl create -f' commands instead.

FATA Error while deploying application: Get http://localhost:8080/api: dial tcp [::1]:8080: connect: connection refused
```

This is my `docker-compose.yml` file's contents
```
[...]

Issue #1201, opened 2019-12-20
Error while deploying application: Get http://localhost:8080/api: dial tcp [::1]:8080: connect: connection refused 


ryannix123:
Latest build of kompose, 1.19.0. This is for an Openshift 3.11 cluster.

hangyan:
This is because most of the early version of kubernetes have 8080 port open ,but not anymore in the latest versions, These settings is hard-coded  in the library kompose ref to ,we are still trying to find some solutions..

hangyan:
Can you try `export KUBERNETES_MASTER=<url>` first?

walter211:
run this in new terminal tab：
`kubectl proxy --port=8080`

glafir:
`└─[0] <> kubectl proxy --port=8080
Starting to serve on 127.0.0.1:8080
`
`export KUBERNETES_MASTER=https://127.0.0.1:8080`
`../bin/./kompose up`

I get this error too (((
`FATA Error while deploying application: Get https://127.0.0.1:6443/api: x509: certificate is valid for 10.96.0.1, 10.50.2.250, not 127.0.0.1`

gargankur2008:
I also have similar issue on the latest kompose . My error is:
`FATA Error while deploying application: Get "https://127.0.0.1:6443/api": dial tcp 127.0.0.1:6443: connect: connection refused `

gargankur2008:
Duplicate of #1131

hangyan:
close by #1297

PR #1204, opened 2019-12-24
Modify default kubernetes client's apiserver url

hangyan:


hangyan:
fix #1131 

#1201 is on openshift, i don't have a openshift env for now, may be check it out later.

hangyan:
@cdrage Now i realize i can force merge a PR ..... I will merge this first, since it's critical for normal usage. You can review it anytime.

PR #1205, opened 2019-12-25
Fix chart lint

### LINKED ISSUE #1200: Helm lint throws errors on chart.yaml
Helm lint command fails on helm files created due to following issues:
- [ERROR] Chart.yaml: directory name (XXX_DIR) and chart name (A new XXX project) must be the same
- [ERROR] Chart.yaml: apiVersion is required. The value must be either "v1" or "v2"

We could fix these issues and contribute to the project. Please let us know if we could create a PR for this.

hangyan:
Fix #1200


hangyan:
Only tested on helm v3

Issue #1202, opened 2019-12-21
doesn't work out of box with the tutorial on the website


kundeng:
A lot of things are outdated? Is this project dead?

kubectl create -f redis-master-deployment.yaml -f redis-master-service.yaml
unable to recognize "redis-master-deployment.yaml": no matches for kind "Deployment" in version "api/v1"

After correcting it, it still doesn't work:
kubectl create -f redis-master-deployment.yaml -f redis-master-service.yaml
error validating "redis-master-deployment.yaml": error validating data: ValidationError(Deployment.spec): missing required field "selector" in io.k8s.api.apps.v1.DeploymentSpec; if you choose to ignore these errors, turn validation off with --validate=false

Even with validate=false, it still doesn't work:
reate -f redis-master-deployment.yaml -f redis-master-service.yaml --validate=false
Error from server (Invalid): error when creating "redis-master-deployment.yaml": Deployment.apps "redis-master" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string{"io.kompose.service":"redis-master"}: `selector` does not match template `labels`]

michielglibert:
See #1180 for fix

hangyan:
I will create a new PR for this these days...So sorry for this inconvenient

Issue #1180, opened 2019-11-13
`kompose up` Does Not Work and Incorrect Deployment Created Via `kompose convert`


loganknecht:
Hello there,

I'm having a significant amount of issues getting `kompose up` and `kompose convert` to work.

# Kubectl Version
```
╭─lknecht@愛と支えボックス ~/Repositories/api_webserver/api_webserver ‹feature/create_minikube_deployment-dev*›
╰─$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
# Kompose Version
```
╭─lknecht@愛と支えボックス ~/Repositories/api_webserver/api_webserver ‹feature/create_minikube_deployment-dev*›
╰─$ kompose version
1.19.0 ()
```

# Docker Compose File
I'm starting with this compose file here
```
version: "3.7"


services:
    api:
        build:
            context: ../.
            dockerfile: ./continuous_integration/api_webserver.dockerfile
        image: rglab/api
        command: ["sh", "-c", "python manage.py makemigrations api &&
                               python manage.py migrate &&
                               gunicorn --bind 0.0.0.0:$$CI_API_PORT api_webserver.wsgi"]
        depends_on:
            - rabbitmq
            - redis
            - postgres
        env_file:
            - ./local_api_docker_compose.env
        networks:
            - backend
        ports:
            - "8000:8000"
        volumes:
            - "api_media:/opt/api_webserver/media"
    worker:
        build:
            context: ../.
            dockerfile: ./continuous_integration/api_webserver.dockerfile
        image: rglab/api
        command: ["sh", "-c", "celery worker --app=api_webserver --loglevel=INFO"]
        depends_on:
            - api
            - rabbitmq
            - redis
            - postgres
        env_file:
            - ./local_api_docker_compose.env
        networks:
            - backend
        volumes:
[...]

PR #1207, opened 2019-12-26
Api version

hangyan:


hangyan:
fix #1202  #1180

hangyan:
This PR use some dirty hacks, since our vendor is locked at a very old versions for all the dependencies. It just works, it's more important for now

PR #1211, opened 2019-12-26
Make env order same as docker-compose

### LINKED ISSUE #1156: kompose env_file superseding environment variables
> kompose version: 1.18.0

Issue: When converting a docker-compose.yml file, if environment variables are specified in both _from_env_ and _environment_, the _environment_ variables are ignored by kompose in favor of the _env_file_ variables.

**Expected behavior:** kompose uses same logic as docker-compose (i.e. the _environment_ variables from docker-compose.yml overwrite those present in _env_file_)

**Actual behavior:** kompose uses variables from _env_file_ regardless of their presence in _environment_

**Steps to reproduce:**

_docker-compose.yml_
```
version: '3.0'
services:
  test:
    image: nginx
    env_file: .env
    environment:
      TEST: from-environment
```

_.env_
```
TEST=from-env_file
```

>$ docker-compose up
>$ docker exec [container] printenv | grep TEST
```
TEST=from-environment
```

>$ kompose convert
```
INFO Kubernetes file "test-deployment.yaml" created
INFO Kubernetes file "test-env-configmap.yaml" created
```

>$ kubectl apply -f test-env-configmap.yaml
>$ kubectl apply -f test-deployment.yaml
>$ kubectl exec [pod] printenv | grep TEST
```
TEST=from-env_file
```

One thing I'm not sure of is whether we would want those _environment_-specified variables to be added as part of the configmap or just as a name/value pair in the spec.  I'm not familiar enough with the kompose project to know what the preferred method is in this case.

hangyan:
fix #1156

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -818,8 +818,17 @@ func (k *Kubernetes) ConfigEnvs(name string, service kobject.ServiceConfig, opt 
 
 	keysFromEnvFile := make(map[string]bool)
 
-	// If there is an env_file, use ConfigMaps and ignore the environment variables
-	// already specified
+	// Load up the environment variables
+	for _, v := range service.Environment {
+
+		envs = append(envs, api.EnvVar{
+			Name:  v.Name,
+			Value: v.Value,
+		})
+
+		keysFromEnvFile[v.Name] = true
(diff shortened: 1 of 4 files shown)


hangyan:
This pr not working , didn't know why the compose envrionment contains envs that defines in the env file

Issue #1186, opened 2019-11-22
Follow docker compose conventions for service fields merge


ChanderG:
According to https://docs.docker.com/compose/extends/, 

1. multi-value options such as `ports`, should be concatenated.
2. Some specific keys should be merged, eg `environment`. 

Currently, in `kompose`, all fields are over-written.

ChanderG:
Previous related issues: #1081

hangyan:
@ChanderG Since there are not so many maintainers by now, if i create a PR, there will be no other guys review it for now. So if you can create a PR for this, that will be great and i will be review it ASAP

ChanderG:
@hangyan Sure, will do so soon.

hangyan:
* Merge
- [] environment

ChanderG:
I have actually already coded out the feature. Have not pushed anything yet as I have to update/write the tests. Will hopefully get around to that by next week.

hangyan:
@ChanderG Great, i will focus on other issues for now

hangyan:
I have merged a lot of PRs these days, you may need to rebase it first..

hangyan:
close by #1226

PR #1226, opened 2020-01-03
concat and merge service fields when merging compose files (#1186)

ChanderG:
As a summary, the following is the break up of the keys specified to be merged/concated:

1. Done and tested
    - Env (merge)
    - Labels (merge)
    - Ports (concat)
    - Volumes (merge by mount dir)
    - Tmpfs (concat) 

2. Done, but not tested/useful (as these are not supported in kompose anyway)
   - DNS (concat)
   - DNS Search (concat)
   - External Links (concat)
   - Expose (concat)

3. Not done as it's not needed
   - Devices (merge)

### DIFF
--- a/pkg/loader/compose/v3.go
+++ b/pkg/loader/compose/v3.go
@@ -564,13 +564,18 @@ func mergeComposeObject(oldCompose *types.Config, newCompose *types.Config) (*ty
 		}
 
 		if len(service.Devices) != 0 {
+			// merge the 2 sets of values
+			// TODO: need to merge the sets based on target values
+			// Not implemented yet as we don't convert devices to k8s anyway
 			tmpOldService.Devices = service.Devices
 		}
 		if len(service.DNS) != 0 {
-			tmpOldService.DNS = service.DNS
+			// concat the 2 sets of values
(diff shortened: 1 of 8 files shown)


ChanderG:
@hangyan ,fyi.

hangyan:
@ChanderG All looks great, thank you very much!

ChanderG:
@hangyan Just noticed that the test I added failed on CI of master, go1.13 build. The same test succeeds on the other 2 builds and on the go1.13 pre-merge PR tests. We may need to investigate.

hangyan:
@ChanderG Yes, i have noticed that, it's very odd. Usually they all fail or succeed. I have use my local kompose done dome convert and checked, it all seems fine. Any ideas?

hangyan:
@ChanderG An occasionally test error on go1.1. gives me some clue, turns out when we merge the volume config, we use a `map`.In golang, the iteration of a map is by random order..So the volumes mount list will not match sometimes.

ChanderG:
@hangyan ,suspected that, did not get time to investigate. Saw your fix. Thanks.

PR #1227, opened 2020-01-04
Support port expose

### LINKED ISSUE #1095: 'expose' is not converted
Using this `docker-compose.yml` for example:
```yml
version: '3'
services:
    seafile:
        container_name: seafile
        image: seafileltd/seafile:latest
        volumes:
            - seafile:/shared
        restart: always
        expose:
            - 80/tcp
```
Does not generate
```yml
  spec:
    ports:
    - name: "80"
      port: 80
      targetPort: 80
```
However, if I change `expose` to `ports`, it does, but it is contradictory to the conversion matrix.

hangyan:
fix #1095


hangyan:
test still failed at `go 1.13`, ignore for now

PR #1255, opened 2020-03-03
Fix printing docker build output if -v is given

fetzerch:
The current version only prints the docker build output only if
the build was successful.


fetzerch:
signed the cla

hangyan:
Looks all good. Thank you very much

PR #1257, opened 2020-03-03
Add support for push-image with --build local

fetzerch:
The PR https://github.com/kubernetes/kompose/pull/1147 introduced a `push-image` parameter for the `up` command that allows to disable pushing the image to the registry. The PR changed the behavior of the `convert` command that now no longer pushes the image.

This commit introduces the same parameter for the `convert` command and reintroduces the original behavior of pushing the image.


hangyan:
Looks all good. Thanks very much

/lgtm

PR #1266, opened 2020-03-15
Add arm64 to make cross

akillcool:
When using aarch64(arm64) machine, we cannot just use the arm binary file directly which may cause "exec format error".


hangyan:
Looks Good.

PR #1272, opened 2020-04-16
incorrect usage of k8s.io/kubernetes

trawler:
When importing `kompose` as a library, I pull in k8s.io/kubernetes as a transitive dependency.
According to https://github.com/kubernetes/kubernetes/issues/81878 any module that has a dependency on k8s.io/kubernetes cannot do minor module upgrades.

This causes the import to fail:
```
➜ go mod why
go: k8s.io/kubernetes@v1.18.1 requires
        k8s.io/api@v0.0.0: reading k8s.io/api/go.mod at revision v0.0.0: unknown revision v0.0.0
```
The recommended way to import these dependencies are by using the published staging repos.

This PR changes the following:
```
k8s.io/kubernetes/pkg/api => k8s.io/api
k8s.io/kubernetes/pkg/apis/extensions => k8s.io/apis/extensions
```


trawler:
closing this PR for further testing!

PR #1273, opened 2020-04-17
Support node.role placement labels

trawler:
This PR adds support for `node.role == manager` and `node.role == worker` placement labels, as described here: https://docs.docker.com/engine/reference/commandline/service_create/#specify-service-constraints-constraint#specify-service-constraints---constraint

The following labels are created:
```
node.role == worker => node-role.kubernetes.io/worker: true
node.role == manager => node-role.kubernetes.io/master: true
```

### DIFF
--- a/docs/conversion.md
+++ b/docs/conversion.md
@@ -31,7 +31,7 @@ __Glossary:__
 | deploy                 | -  | -  | ✓  |                                                             |                                                                                                                |
 | deploy: mode           | -  | -  | ✓  |                                                             |                                                                                                                |
 | deploy: replicas       | -  | -  | ✓  | Deployment.Spec.Replicas / DeploymentConfig.Spec.Replicas   |                                                                                                                |
-| deploy: placement      | -  | -  | ✓  | Pod.Spec.NodeSelector                                       |                                                                                                                |
+| deploy: placement      | -  | -  | ✓  | Pod.Spec.NodeSelector                                       |    Supported constraints: `node.hostname`, `node.role == worker`, `node.role == manager`, `engine.labels.operatingsystem` and `node.labels.xxx`                                                                                                         |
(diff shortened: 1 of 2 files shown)


cdrage:
Thanks @trawler ! Only thing left is documentation, do you mind updating the table of supported labels, etc?

https://github.com/kubernetes/kompose/blob/master/docs/conversion.md

and the label section here: https://github.com/kubernetes/kompose/blob/master/docs/user-guide.md

trawler:
hi @cdrage, thanks for taking the time to look into my PR.
I updated the `docs/conversion.md` file, but I left out the user-guide labels section because I believe it is only referring to kompose specific labels. 
Let me know If I misunderstood.

hangyan:
I have checked some living kubernetes cluster ,the role annotations seems to be:
image

image

Env: 
Kubernetes 1.16

trawler:
> I have checked some living kubernetes cluster ,the role annotations seems to be:
> image
> 
> image
> 
> Env:
> Kubernetes 1.16

Hi @hangyan! that's strange... can you share the original placement labels from the compose file you used for the conversion?

hangyan:
> > I have checked some living kubernetes cluster ,the role annotations seems to be:
> > image
> > image
> > Env:
> > Kubernetes 1.16
> 
> Hi @hangyan! that's strange... can you share the original placement labels from the compose file you used for the conversion?
> 
> I have used "./kompose convert" to test this PR, and found that
> Screenshot 2020-04-22 at 18 10 46
> 
> was succesfuly converted to
> Screenshot 2020-04-22 at 18 10 23

I didn't mean the transation was wrong. I mean, If you tranlate this labels and there values are `true`. But in the kubernetes cluster, the values are "", whouldn't that a mismatch?

PR #1279, opened 2020-05-25
Update conversion for command and entrypoint

xavierleune:
This pull requests updates the documentation according to the current behavior of kompose (tested with 1.19.0)


xavierleune:
Checking CLA

Issue #1275, opened 2020-04-23
Deprecating kompose up and kompose down in favor of only kompose convert


cdrage:
So... a few issues:

1. We're using an old library and it's very hard / a LOT of changes in order to go from the old golang kubernetes library to the new one. (ex: https://github.com/kubernetes/kompose/issues/1264)
2. Lot's of issues with regards to connecting to local servers (minikube isn't port 8080 anymore..)
3. `kompose convert` is *very* popular in terms of converting. Most people using `kompose convert -o foobar && kubectl apply -f foobar/`
4. We do not have enough maintainers to maintain using `kompose up` and `kompose down`, using only `kompose convert` would reduce the amount of work and maintainability will be easier.

Should we just deprecate `kompose down` and `kompose up`?

Ping @kubernetes/kompose-admins @kubernetes/kompose-maintainers @kadel @hangyan @surajnarwade @surajssd @hangyan @ngtuna

surajnarwade:
I agree , +1 from my side

trawler:
Since we started exploring Kompose I can only agree with the above. `kompose up/down` was never an option for us, but we love the convert + helm functionality.
So +1  👍

cdrage:
Thanks all, even though it's only been 24 hours, I've got nothing but praise in the e-mail as well as this issue.

I'll change this in **one week** in order to give a **lazy consensus** in terms of removing `kompose up` and `kompose down`.

hangyan:
+1

kaceo:
New to Kubernetes stack, so my opinion is only from a docker-compose user.  Is it correct to say the tool currently only handles partial conversion?  I am thinking about the use of persistent volume, ingress, depends on, secrets, environment vars, etc. these need further tweaking before sending to kubectl?  If so, it is probably better to make this tool only focus on convert, and add hints and lints to help the compose user fill in the missing gaps.  As a new Kubernetes user I am most interested to convert a fully working docker app into Kubernetes with least amount of re-architecting, and so far this tool has helped me but still left a lot of the puzzle.

gargankur2008:
I agree. My `kompose convert` created about 15 yaml files, is there a way to apply all of them without putting them in terminal ?

cdrage:
Sorry for the delay, I'm looking at implementing this shortly.

gargankur2008:
> Sorry for the delay, I'm looking at implementing this shortly.

[...]

Issue #1293, opened 2020-07-26
Kompose down fails due to a failed connection to https://127.0.0.1:6443/api


guyarb:
I used `kompose up` and got `FATA Error while deleting application: Get https://127.0.0.1:6443/api: dial tcp 127.0.0.1:6443: connect: connection refused`, after short lookup in google, I've found the flag `--server`.
When I tried using `kompose down` I got the same error, but I didn't see the same flag in `kompose down -h`.
I'm using kompose of version `1.21.0 (992df58d8)`  on ubuntu 18.04 machine.

fullkomnun:
I had the same issue, try to check if that is the port for the API server in '~/.kube/config'
and if necessary use the '--server' flag to explicitly specify it

hangyan:
close by #1297.

Issue #1284, opened 2020-06-26
Kompose up with remote Openshift


vertti:
I have a Openshift cluster running (not locally). My kubectl connects to it correctly. But running 

```
kompose --provider=openshift --volumes emptyDir --build build-config up
```
fails to connect to localhost, where it definitely shouldn't be connecting in the first place:

```
FATA Error while deploying application: Get https://127.0.0.1:6443/api: dial tcp 127.0.0.1:6443: connect: connection refused
```

Does everyone just use `kompose` with local minikube etc, or what am I doing wrong?

hangyan:
close by #1297

Issue #1278, opened 2020-05-20
kompose up seems to not use the build context's .dockerignore file


michaeldbianchi:
When attempting to run `kompose up`, I run into a build error referenced in issue [656](https://github.com/kubernetes/kompose/issues/656) saying:

```
FATA Error while deploying application: k.Transform failed: Unable to build Docker image for service wrangler: Unable to create a tarball: read /Users/michaelbianchi/workspace/mercury-platform/wrangler/_build/dev/lib/certifi/ebin: is a directory
```

While the other issue states `kompose` doesn't handle symlinks well, this directory should never be included in the build context due to being ignored in the `.dockerignore` file. If `kompose` is going to handle the building of images, it should respect things handled by `docker-compose` such as `.dockerignore`.

It's also possible I have this poorly setup, but everything else looks right. `kompose convert` works. Source code is held in https://github.com/michaeldbianchi/mercury-platform

Verbose:
```
 λ on minikube in ~/workspace/mercury-platform on  master [!?] kompose -f docker-compose.yml -v up
DEBU Docker Compose version: 3
WARN Volume mount on the host "/Users/michaelbianchi/workspace/mercury-platform/data" isn't supported - ignoring path on the host
INFO Build key detected. Attempting to build image 'michaeldbianchi/wrangler:dev'
DEBU Build image working dir is: /Users/michaelbianchi/workspace/mercury-platform
DEBU Build image context is: /Users/michaelbianchi/workspace/mercury-platform/wrangler
INFO Building image 'michaeldbianchi/wrangler:dev' from directory 'wrangler'
DEBU Created temporary file /var/folders/65/6cd8mczj1d198sxvrq2vb7v40000gp/T/kompose-image-build-475960662 for Docker image tarballing
FATA Error while deploying application: k.Transform failed: Unable to build Docker image for service wrangler: Unable to create a tarball: read /Users/michaelbianchi/workspace/mercury-platform/wrangler/_build/dev/lib/certifi/ebin: is a directory
```

michaeldbianchi:
Note that if I delete the _build and deps directories (which are in the .dockerignore), `kompose up` worked just fine. However, since elixir builds creates directories with symlinks, it makes kompose almost unusable since every time I run tests or a dev server, those directories are populated.

hangyan:
close by #1297

PR #1297, opened 2020-07-31
Remove up and down cmd

hangyan:


hangyan:
fix #1275 #1293 #1284 #1278 #1277 #1264 #1238 #1201

PR #1300, opened 2020-08-01
fix stop_grace_period in v3

jing2uo:
fix [issue 1298](https://github.com/kubernetes/kompose/issues/1298).

### DIFF
--- a/pkg/loader/compose/v3.go
+++ b/pkg/loader/compose/v3.go
@@ -17,10 +17,11 @@ limitations under the License.
 package compose
 
 import (
-	"github.com/spf13/cast"
 	"strconv"
 	"strings"
 	"time"
+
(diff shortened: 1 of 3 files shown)


jing2uo:
I already signed cla ...

jing2uo:
I'm so sorry for leaving a bunch of failed records in the ci build history..

pkg/loader/compose/utils.go
```diff
@@ -144,6 +145,12 @@ func normalizeNetworkNames(netName string) (string, error) {
 	return netval, nil
 }
 
+// convert *time.Duration to String
+func DurationToString(d *time.Duration) string {
```
hangyan: seems no need to create a function

PR #1343, opened 2020-11-03
make kompose exit with 1 if error

namusyaka:
Old code passes -1 to os.Exit and the value is directly passed to syscall.
The behavior with -1 depends on environments (most cases should convert
into 255 though)

This commit removes unnecessary the ambiguity and changes it to use 1 instead of 255.

---
Also, this change is in favor of https://golang.org/pkg/os/#Exit:

```
For portability, the status code should be in the range [0, 125].
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,8 +16,15 @@ limitations under the License.
 
 package main
 
-import "github.com/kubernetes/kompose/cmd"
+import (
+	"log"
+
+	"github.com/kubernetes/kompose/cmd"
+)
 
 func main() {
-	cmd.Execute()
+	if err := cmd.Execute(); err != nil {
+		log.SetFlags(0)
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -16,8 +16,15 @@ limitations under the License.
...
+
+	"github.com/kubernetes/kompose/cmd"
+)
 
 func main() {
-	cmd.Execute()
+	if err := cmd.Execute(); err != nil {
+		log.SetFlags(0)
+		log.Fatal(err)
+	}
```
namusyaka, line 29: I think the error raised by the cmd package should be handled by main.go.
But some `log.Fatalf` still exist inside the package. I think those logging can be extracted here.

hangyan:
looks great. Thanks

PR #1344, opened 2020-11-03
Follow up on #1341

namusyaka:
Just following up; Please ignore if it does not make sense.


hangyan:
@namusyaka Thanks, this was some testing-purpose log, it shouldn't be commit to github....

namusyaka:
@hangyan Ah ok, should we remove it? Although I rephrased the message in this PR, I agreed that removing the logs.

hangyan:
yes, remove it is better

PR #1345, opened 2020-11-03
Support port range for docker-compose v2

namusyaka:
In docker-compose v2, the port range syntax is available: https://docs.docker.com/compose/compose-file/compose-file-v2/#ports

However, kompose does not support this. I supposed that the v1/v2 loader can't use docker's parsing mechanism like v3 smoothly because v2 loader needs to keep host IP support.

Considering above, I was trying to make kompose support the port range in docker-compose v2 by using docker's opts and nat package partially.
If you intentionally don't support port range in v1/v2, please let me know.

Also, I rewrote some tests for loadPorts, this contains a few fixes such as invalid logs.


hangyan:
This looks great. It is always welcome to mapping more compose spec in kompose even we are now manily on maintaince mode. Therre are more test need to be added though

`script/test/fixtures/v2/docker-compose.yaml` : this file is intended to act as a full example of v2 compose, a script in `pkg/test/cmd/tests_new.sh` will convert this yaml and compare with the output both in kubernetes and openshift provider. Feel free to add more services or modify the existing service in the compose yaml file to test this new port-range feature

namusyaka:
@hangyan Thanks for taking a look at this. I added tests to tests according to your advice and fixed failing tests related to #1341

PR #1348, opened 2020-11-04
cleanup debug logs & remove duplicated output

namusyaka:
/cc @hangyan

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -49,14 +49,14 @@ var (
 
 // RootCmd root level flags and commands
 var RootCmd = &cobra.Command{
-	Use:   "kompose",
-	Short: "A tool helping Docker Compose users move to Kubernetes",
-	Long:  `Kompose is a tool to help users who are familiar with docker-compose move to Kubernetes.`,
+	Use:           "kompose",
+	Short:         "A tool helping Docker Compose users move to Kubernetes",
+	Long:          `Kompose is a tool to help users who are familiar with docker-compose move to Kubernetes.`,
+	SilenceErrors: true,
 	// PersistentPreRun will be "inherited" by all children and ran before *every* command unless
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -17,14 +17,15 @@ limitations under the License.
...
 
 	"github.com/kubernetes/kompose/cmd"
 )
 
 func main() {
 	if err := cmd.Execute(); err != nil {
-		log.SetFlags(0)
-		log.Fatal(err)
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
```
namusyaka, line 29: @hangyan I'm sorry, I changed the code to avoid touching global state of the log package.

cmd/root.go
```diff
@@ -49,14 +49,14 @@ var (
 
 // RootCmd root level flags and commands
 var RootCmd = &cobra.Command{
-	Use:   "kompose",
-	Short: "A tool helping Docker Compose users move to Kubernetes",
-	Long:  `Kompose is a tool to help users who are familiar with docker-compose move to Kubernetes.`,
+	Use:           "kompose",
+	Short:         "A tool helping Docker Compose users move to Kubernetes",
+	Long:          `Kompose is a tool to help users who are familiar with docker-compose move to Kubernetes.`,
+	SilenceErrors: true,
```
namusyaka, line 55: I would like to avoid duplicated outputs by enabling `SilenceErrors`.
Does this make sense?

Ref: there are two places to be used:
- https://github.com/spf13/cobra/blob/86f8bfd7fef868a174e1b606783bd7f5c82ddf8f/command.go#L940
- https://github.com/spf13/cobra/blob/86f8bfd7fef868a174e1b606783bd7f5c82ddf8f/command.go#L969

hangyan (approved):

PR #1351, opened 2020-11-06
retrieve container test by updating Dockerfile

namusyaka:
`make test-image` and `make test-container` seem broken, is there any chance to recover these tasks?


hangyan:
I thinks we can add it to the `.travis.yml` file.

namusyaka:
Thanks for the review, @hangyan !
Since the current entrypoint of the Dockerfile runs [`make test`](https://github.com/kubernetes/kompose/blob/01f18cb3cf3f081b33c057f602e5f22345afb808/script/test_in_container/run.sh#L25), I don't think we need to add it to the travis integration.
I was thinking recovering the container-test is useful to run other tests on the container by overwriting `--entrypoint` with something like `/bin/bash`. WDYT?

hangyan:
seems not much test we have to put in the container now.

hangyan:
I'm gonna merge this for now. If any good test idea come up later, please open a new PR

hangyan (approved):

namusyaka:
Will do, thanks @hangyan !

PR #1373, opened 2021-03-25
add support for golangci-lint

remyleone:


hangyan:
The lint ci failed

remyleone:
@hangyan fixed. I've commented on some linters that would take a bit of time to be fixed. Feel free to explore them and judge whether or not they are worth it.

remyleone:
You can also check all the available linters with `golangci-lint linters` and see if you want to include new ones.

hangyan:
I think the current set is good enough. There is no need to consider all sub linters at first, we can work on it later

hangyan (approved):

Issue #1359, opened 2020-11-30
Panic when converting Docker compose service with 'start_first' to kubernetes


Johan-Lindvall-Stratsys:
Given the following service definition
```
version: '3.5'

services:
  foo:
    image: image
    deploy:
      update_config:
        order: start-first
```
Kompose (version 1.22, locally built), crashes:
```
kompose.exe convert -f=crash.yml
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x424d09]

goroutine 1 [running]:
k8s.io/apimachinery/pkg/util/intstr.(*IntOrString).String(0x0, 0xc0001e0250, 0xf39080)
        C:/Users/johan.lindvall/go/pkg/mod/k8s.io/apimachinery@v0.19.0-rc.2/pkg/util/intstr/intstr.go:93 +0x29
github.com/kubernetes/kompose/pkg/transformer/kubernetes.(*Kubernetes).InitD(0xc000178000, 0xc00008ec38, 0x3, 0x0, 0x0, 0xc0003c5510, 0x5, 0x0, 0x0, 0x0, ...)
        C:/source/kompose/pkg/transformer/kubernetes/kubernetes.go:382 +0x3fd
github.com/kubernetes/kompose/pkg/transformer/kubernetes.(*Kubernetes).CreateKubernetesObjects(0xc000178000, 0xc00008ec38, 0x3, 0x0, 0x0, 0xc0003c5510, 0x5, 0x0, 0x0, 0x0, ...)
        C:/source/kompose/pkg/transformer/kubernetes/kubernetes.go:1020 +0x1c8
github.com/kubernetes/kompose/pkg/transformer/kubernetes.(*Kubernetes).Transform(0xc000178000, 0xc0002b9500, 0xf2d354, 0x7, 0xc0002b91d0, 0x100000100, 0x0, 0x0, 0x0, 0x0, ...)
        C:/source/kompose/pkg/transformer/kubernetes/kubernetes.go:1165 +0x11a5
github.com/kubernetes/kompose/pkg/app.Convert(0x100000100, 0x0, 0x0, 0x0, 0x0, 0xf2b6e3, 0x4, 0x1, 0xf3cec4, 0x15, ...)
        C:/source/kompose/pkg/app/app.go:233 +0x282
github.com/kubernetes/kompose/cmd.glob..func3(0x1747be0, 0xc0003eec60, 0x0, 0x1)
[...]

PR #1378, opened 2021-04-30
Fix convert deployment log error

hangyan:


hangyan:
Fix #1359 

All credits to @Johan-Lindvall-Stratsys

PR #1393, opened 2021-06-27
actually use normalizeNetworkNames to produce valid network names

exokortex:
Currently kompose does not normalize network names, which can lead to networks containing invalid characters for Kubernetes NetworkPolicy resources. There already was a function for this in ` pkg/loader/compose/utils.go`, but it was not used.


hangyan:
not related go module error, will fix it after merged this PR

hangyan:
Thanks @exokortex

PR #1405, opened 2021-08-06
fix: --build flag default change to local

### LINKED ISSUE #1363: --build should not automatically push
I'm concerned that the default configuration for `kompose` if you want kompose to build your images for you is to attempt to push company intellectual property to a public server.

Is it worth having a breaking change that defaults to not push?

Folks may be well-versed in contributing open source containers and have the configuration to deploy to dockerhub. Then they want to learn kubernetes and transfer their knowledge. They come across https://kubernetes.io/docs/tasks/configure-pod-container/translate-compose-kubernetes/ and see they can get their company's private docker repos running locally on kubernetes pretty easily. They follow the instructions and run `kompose convert --build local` just like in the docs to build locally and their intellectual property goes to a public server.

The kompose documentation says you can use --build to build and push your containers. But despite that I don't think anyone expects the default functionality of a "build" step is to push it to a public server.

Thankfully I don't have credentials to push to docker.io :)

zeromake:
Fix #1363

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -143,8 +143,8 @@ func init() {
 	convertCmd.Flags().MarkHidden("build-branch")
 
 	// Standard between the two
-	convertCmd.Flags().StringVar(&ConvertBuild, "build", "local", `Set the type of build ("local"|"build-config"(OpenShift only)|"none")`)
-	convertCmd.Flags().BoolVar(&ConvertPushImage, "push-image", true, "If we should push the docker image we built")
+	convertCmd.Flags().StringVar(&ConvertBuild, "build", "none", `Set the type of build ("local"|"build-config"(OpenShift only)|"none")`)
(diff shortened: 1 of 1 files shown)


hangyan:
@zeromake Please follow the instructions to sign the CLA first

zeromake:
@hangyan just finished

hangyan:
@zeromake I read the related issues, and it seems it's better to change the push-image flag's default? not the build flag?

zeromake:
@hangyan 
>by google translate

After thinking about it, modifying the default value of the push-image option would be a better choice.

hangyan:
@zeromake  You may need also go through the user docs to update related sections.

zeromake:
@hangyan 
I looked through the document and did not find the content related to these two options

hangyan:
@zeromake  OK, I will review the issue and PR to see if we can just merge this.

hangyan (approved):

PR #1419, opened 2021-08-24
Add support for negated placement constraint

### LINKED ISSUE #1132: Support for negated placement constraint docker-compose v3
Code currently only allows for ' == ' constraint, while specs also allow ' != ', also the whitespace surrounding the comparison operator does not seem to be mandatory.

https://github.com/kubernetes/kompose/blob/v1.18.0/pkg/loader/compose/v3.go#L138

https://docs.docker.com/engine/reference/commandline/service_create/#specify-service-constraints---constraint

ichxxx:
close #1132


pkg/kobject/kobject.go
```diff
@@ -137,16 +137,16 @@ type ServiceConfig struct {
...
+	TmpFs              []string                             `compose:"tmpfs"`
+	Dockerfile         string                               `compose:"dockerfile"`
+	Replicas           int                                  `compose:"replicas"`
+	GroupAdd           []int64                              `compose:"group_add"`
+	Volumes            []Volumes                            `compose:""`
 	Secrets            []dockerCliTypes.ServiceSecretConfig
-	HealthChecks       HealthChecks      `compose:""`
-	Placement          map[string]string `compose:""`
+	HealthChecks       HealthChecks                         `compose:""`
+	Placement          []corev1.NodeSelectorRequirement     `compose:""`
```
hangyan, line 149: The `ServiceConfig` struct is more similar to docker-compose data structure. By far we haven't introduce any kubernetes-specific fields into it. So i think we should keep it that way for now. You can define a custom struct to represent the placement requirement of the docker-compse and parse it in kubernetes package

pkg/loader/compose/compose_test.go
```diff
@@ -502,9 +502,26 @@ func TestCheckPlacementCustomLabels(t *testing.T) {
...
+			t.Errorf("operator is not equal, expected %s, got %s", expected[i].Operator, output[i].Operator)
+		}
+		if len(output[i].Values) != len(expected[i].Values) {
+			t.Errorf("values len is not equal, expected %d, got %d", len(expected[i].Values), len(output[i].Values) )
+		}
+		if len(output[i].Values) > 0 && output[i].Values[0] != expected[i].Values[0] {
+			t.Errorf("value is not equal, expected %s, got %s", expected[i].Values, output[i].Values)
+		}
 	}
 }
```
hangyan, line 527: A docker-compose file test is also needed .Please refer to `make test-cmd` to checkout how it works

pkg/kobject/kobject.go
ichxxx: That's a better way

pkg/loader/compose/compose_test.go
ichxxx: Sorry, I missed it, test has been updated.

pkg/kobject/kobject.go
```diff
@@ -203,6 +203,18 @@ type Volumes struct {
...
 
+// Placement holds the placement struct of container
+type Placement struct {
+	Constraints []Constraint
+}
+
+// Constraint holds the constraint struct in the Placement of the container
+type Constraint struct {
+	Key      string
+	Operator corev1.NodeSelectorOperator
```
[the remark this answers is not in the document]
hangyan: Sure

PR #1420, opened 2021-08-25
Add golangci-lint

aiyijing:
Signed-off-by: aiyijing <aiyijing@live.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,3 +117,7 @@ test-container:
 .PHONY: test-k8s
 test-k8s:
 	./script/test_k8s/test.sh
+
+.PHONY: golangci-lint
+golangci-lint:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -117,3 +117,7 @@ test-container:
 .PHONY: test-k8s
 test-k8s:
 	./script/test_k8s/test.sh
+
+.PHONY: golangci-lint
+golangci-lint:
+	golangci-lint run -c .golangci.yml
```
hangyan, line 123: 1. Do we have a `.golangci.yml`? Does this behaviour consistent with github ci?
2. If golangci-lint has not been installed yet, may be we can provide a go get command in the setup part to install it

aiyijing: That's a good idea

PR #1425, opened 2021-08-30
Add support for placement preferences docker-compose v3.3+

ichxxx:
https://docs.docker.com/engine/reference/commandline/service_create/#specify-service-placement-preferences---placement-pref

https://docs.docker.com/engine/swarm/services/#placement-preferences

> Placement preferences try to place tasks on appropriate nodes in an algorithmic way (currently, only spread evenly).
> Placement preferences are not strictly enforced. If no node has the label you specify in your preference, the service is deployed as though the preference were not set.

So, I think we can use affinity to support it.
For exmple:
```yaml
deploy:
  placement:
    preferences:
      - spread: node.labels.ssd
```
can be converted to:
```yaml
affinity:
  nodeAffinity:
    preferredDuringSchedulingIgnoredDuringExecution:
      - preference:
          matchExpressions:
            - key: ssd
              operator: Exists
        weight: 1

### DIFF
--- a/pkg/kobject/kobject.go
+++ b/pkg/kobject/kobject.go
@@ -215,6 +215,7 @@ type Volumes struct {
 type Placement struct {
 	PositiveConstraints map[string]string
 	NegativeConstraints map[string]string
+	Preferences         []string
 }
 
 // GetConfigMapKeyFromMeta ...
--- a/pkg/loader/compose/compose_test.go
+++ b/pkg/loader/compose/compose_test.go
@@ -499,6 +499,11 @@ func TestCheckPlacementCustomLabels(t *testing.T) {
 			"node.labels.something == anything",
 			"node.labels.monitor != xxx",
 		},
+		Preferences: []types.PlacementPreferences{
+			{Spread: "node.labels.zone"},
+			{Spread: "foo"},
+			{Spread: "node.labels.ssd"},
+		},
 	}
 	output := loadV3Placement(placement)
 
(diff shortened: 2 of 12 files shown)


hangyan:
@ichxxx Hi, Thanks for the PR. I think affinity is not the way to do it. Check the topology feature of kubernetes, I think may be this is the way: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/

ichxxx:
@hangyan OK, I will check that.

hangyan:
1. https://kubernetes.io/docs/reference/labels-annotations-taints/ Please check this document. I think which topoloy key we can convert to still need some research. The current solution may not the best one.
2. The current e2e test script is `test_new.sh`, many old fixured won't be tested by now. This is due to some historical reasons. Can you add you updated fixtures to the new script? I think the current github actions does not tests this part.

ichxxx:
> 1. https://kubernetes.io/docs/reference/labels-annotations-taints/ Please check this document. I think which topoloy key we can convert to still need some research. The current solution may not the best one.
> 2. The current e2e test script is `test_new.sh`, many old fixured won't be tested by now. This is due to some historical reasons. Can you add you updated fixtures to the new script? I think the current github actions does not tests this part.

1. I think the topology key we can convert in placement preferences is the same as placement constraints. This has been updated. And the label `node.labels.xxx` is a user-defined label, I think just trim the prefix `node.labels.` is well.
2. Test has already updated. I reuse the test of `fixtures/deploy/placement`. Shoud I separate it?

hangyan:
I think the current solution should works fine, no further updated needed.

hangyan (approved):

PR #1441, opened 2021-10-12
Fix typo in empty volume warning

sbs2001:


hangyan:
Seem there are missing test about empty volumes. You can add some unittest or cmd test.

PR #1455, opened 2021-11-06
Add make rule for darwin arm64 in cross target

sbs2001:
Addresses https://github.com/kubernetes/kompose/issues/1416


hangyan (approved):

hangyan:
Looks good.

PR #1456, opened 2021-11-07
Infer storage type using label

sbs2001:
This allows users to set volume type individually for each service by providing the 
`kompose.volume.type` label. See cmd test for examples.

Fixes https://github.com/kubernetes/kompose/issues/1400


hangyan:
Loogs good for now. I will review this later. Plan to relase 1.26 after merge this

PR #1478, opened 2022-01-19
Fixed secret file locations to match results from docker-compose when using file-based secrets

### LINKED ISSUE #1280: kompose secrets file location
It appears there is a mismatch between where secrets are placed when using kompose vs docker compose

docker compose places secrets in the folder `/run/secrets` kompose defines the mount path for secrets in the folder `/run/secrets/{secret-name}`. This is a breaking change if you are using `kompose convert` to deploy to k8s and need both docker-compose and k8s to work

campos-ddc:
Fixes #1280

--

Still need to do some tests to make sure everything works, hopefully CI will also point out any issues.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -63,6 +63,12 @@ var (
 
 	ServiceGroupMode string
 	ServiceGroupName string
+
+	// SecretsAsFiles forces secrets to result in files inside a container instead of symlinked directories containing
+	// files of the same name. This reproduces the behavior of file-based secrets in docker-compose and should probably
+	// be the default for kompose, but we must keep compatibility with the previous behavior.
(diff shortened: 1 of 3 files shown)


campos-ddc:
Tested with a local k8s cluster and files seem to be in the right place when using the new flag (replicating docker-compose behavior).

campos-ddc:
The original code came from https://github.com/kubernetes/kompose/pull/1159, looks like @hangyan reviewed that one.

campos-ddc:
Any chance that this will ever get reviewed?

I'm really trying here, and would hate to give up and live with a fork forever.

kadel:
/lgtm

Thank you, @campos-ddc for contributing to kompose.
Sorry that it took a little bit longer 😇

PR #1486, opened 2022-03-12
Add kompose.service.expose.ingress-class-name

sergeyshevch:
I've added support for customizing ingressClassName in kubernetes ingress resource with label `kompose.service.expose.ingress-class-name`.


sebgoa:
I have not worked with kompose for many years so it is better to unassign myself.

sergeyshevch:
@cdrage @hangyan I also asked you in a K8s slack. Maybe one of you can look at it?

PR #1490, opened 2022-03-21
fix(v3): Added network name normalisation in v3 parsing

### LINKED ISSUE #1487: can not convert `_` of `Network` to `-`
When I was use `kompose convert -f docker-compose.yaml -c` to convert compose file to chart. My compose file `Network` contains `_` like `test_network`, and `kompose` just convert to `test_network` service. But `_` is invalid character in kubernetes service.

May be we should make it more intelligent, a simple way to do this is to convert `_` to `-`.

screenshoot:

image

### If this is necessary, I am happy to do this part of the coding.

sergeyshevch:
Fixes #1487 
Current parsing of network names in v3 doesn't provide network name normalization that causes incorrect label names and incorrect networkpolicy resource name


sergeyshevch:
Also i provided updates in testcase for v3 version that checks that now normalization works correctly

PR #1512, opened 2022-06-30
fix daemonset creation (add selector)

realgam3:
To fix daemonset creation, add a selector


realgam3:
**A dreamonset created with selector=null causes an error when applied:**
`error: error validating "test-daemonset.yaml": error validating data: ValidationError(DaemonSet.spec): missing required field "selector" in io.k8s.api.apps.v1.DaemonSetSpec; if you choose to ignore these errors, turn validation off with --validate=false`

realgam3:
Hi @kadel and @surajssd,
Can you review this pull request?
It fix an error I encounter lately

PR #1515, opened 2022-08-16
fix: change the port ID definition in kubernetes.go file

AhmedGrati:
# Problem
Fix this [issue](https://github.com/kubernetes/kompose/issues/1514). The problem was mainly, in the ID of each port.
In the previous implementation, the ID was composed of the ASCII code of the port number and its protocol. However, such implementation led to some inconveniences for the large port numbers. In fact, If you try this command `fmt.Println(string(55546) == string(55547))` it will print `true`, which is not the desired behavior and it's the main cause behind the issue.
# Solution
To overcome this issue, I just replaced the `string()` function with `strconv.Itoa()`. I added large close numbers in the tests suite, to ensure that we get the desired behavior.

### DIFF
--- a/pkg/kobject/kobject.go
+++ b/pkg/kobject/kobject.go
@@ -18,6 +18,7 @@ package kobject
 
 import (
 	"path/filepath"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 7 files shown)


cdrage:
LGTM!

PR #1521, opened 2022-09-25
fix: update the container name function to format the returned name

AhmedGrati:
# Problem
This pull request aims to solve [this issue](https://github.com/kubernetes/kompose/issues/1519).
The problem was in the `GetContainerName` function located in the `pkg/transformer/kubernetes/k8sutils.go` file.
In fact, in the previous implementation if the container name of the service is empty we return the service name without formatting it, by replacing `_` with `-`, which results in creating containers with names that contain `_`.

# Solution
To overcome this issue, I just updated the function `GetContainerName` to format the returned name. I added in the unit test, a test function dedicated to check the desired behavior.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -37,10 +37,10 @@ Download from [GitHub](https://github.com/kubernetes/kompose/releases/download/v
 
 ## Go
 
-Installing using `go get` pulls from the master branch with the latest development changes.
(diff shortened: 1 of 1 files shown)


AhmedGrati:
Re-opened [here](https://github.com/kubernetes/kompose/pull/1528).

Issue #1539, opened 2022-11-21
Option to output conversion to one file.


cdrage:
Similar to helm and other k8s tools, we should be able to have the option to output to one file, rather than multiple files.

Or maybe we should make it the default that everything is outputted to one file? What do people think?

AhmedGrati:
@OneCricketeer For debugging purposes, you might occasionally need to display all the manifests in one file so that you don't have to switch between them when a problem arises. Therefore, IMO I see that the default behavior should remain the same, and we add the `--one-file` flag to save the output in one file.

AhmedGrati:
After investigating this issue, I found that this functionality already exists, by using the `--file` flag. For example:
```sh
kompose convert -f docker-compose.yaml --out f.yaml
```
But the issue is that the output is not the correct one. In fact, for this docker-compose file:
```yaml
version: "3.8"

services:
  front_end:
    image: gcr.io/google-samples/gb-frontend:v4
    ports:
      - "80:80"
    environment:
      - GET_HOSTS_FROM=dns
    labels:
      kompose.service.expose: lb
      kompose.service.expose.ingress-class-name: nginx
```
we'll have the following output:
```yaml
apiVersion: v1
items:
  - apiVersion: v1
    kind: Service
    metadata:
      annotations:
        kompose.cmd: /var/folders/jn/kphqd_553871ydfcs_2_5kbm0000gn/T/go-build1291552472/b001/exe/main convert -f docker-compose.yaml --out f.yaml
        kompose.service.expose: lb
        kompose.service.expose.ingress-class-name: nginx
        kompose.version: 1.26.1 (HEAD)
      creationTimestamp: null
      labels:
        io.kompose.service: front-end
      name: front_end
    spec:
      ports:
        - name: "80"
          port: 80
          targetPort: 80
      selector:
        io.kompose.service: front-end
    status:
      loadBalancer: {}
  - apiVersion: apps/v1
    kind: Deployment
    metadata:
      annotations:
        kompose.cmd: /var/folders/jn/kphqd_553871ydfcs_2_5kbm0000gn/T/go-build1291552472/b001/exe/main convert -f docker-compose.yaml --out f.yaml
        kompose.service.expose: lb
        kompose.service.expose.ingress-class-name: nginx
        kompose.version: 1.26.1 (HEAD)
      creationTimestamp: null
      labels:
        io.kompose.service: front-end
      name: front-end
    spec:
      replicas: 1
      selector:
        matchLabels:
[...]

PR #1541, opened 2022-11-28
Fix: Make the out flag print to one file using seperator

AhmedGrati:
# Description
This pull request solves #1539.

# Implementation
Basically, it has the exact implementation of the `helm template` command. It will output all the k8s resources in one YAML file using the `---` separator. If the user specifies a JSON output, it will return an error since we cannot have multiple JSON resources in one file.


pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -148,16 +148,16 @@ func getDirName(opt kobject.ConvertOptions) string {
...
 
-	bytes, _ := json.Marshal(object)
-	r.Raw = bytes
+// 	bytes, _ := json.Marshal(object)
+// 	r.Raw = bytes
 
-	return r
-}
+// 	return r
+// }
```
cdrage, line 160: remove commented out section

pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -192,36 +192,30 @@ func PrintList(objects []runtime.Object, opt kobject.ConvertOptions) error {
...
-		}
-		printVal, err := transformer.Print("", dirName, "", data, opt.ToStdout, opt.GenerateJSON, f, opt.Provider)
-		if err != nil {
-			return errors.Wrap(err, "transformer.Print failed")
+			data, err := marshal(versionedObject, opt.GenerateJSON, opt.YAMLIndent)
+			if err != nil {
+				return fmt.Errorf("error in marshalling the List: %v", err)
+			}
+			data = []byte(fmt.Sprintf("---\n%s", data))
+			printVal, err := transformer.Print("", dirName, "", data, opt.ToStdout, opt.GenerateJSON, f, opt.Provider)
```
cdrage, line 213: good, but you need to add comments "this part add `---` which unifies the file", etc.

pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -192,36 +192,30 @@ func PrintList(objects []runtime.Object, opt kobject.ConvertOptions) error {
...
+			if err != nil {
+				return errors.Wrap(err, "transformer.Print failed")
```
cdrage, line 215: I know this is outside the scope of the PR, but perhaps change this error output while you're at it :) "transformer to print to one single file failed" or something like that.

cdrage (approved):
boom-mind-blown

SO HAPPY THIS MAKES TESTS PASS.

Thank you! Small minor changes, but otherwise code LGTM.

Update the PR with the changes and then feel free to merge.

hangyan (approved):

hangyan:
LGTM

PR #1550, opened 2023-01-13
Fix CI

cdrage:
Fix CI

Accidently removed the check-gofmt.sh script.


Lctrs:
I think this whole script could be just replaced by this Makefile target : 
```
.PHONY: gofmt
gofmt:
	gofmt -l ./
```

Lctrs:
Nevermind, it never exits with a status other than `0`.

cdrage:
Merging PR as I've fixed the issues! I've also removed SemaphoreCi (we haven't used it for yearssss now).

PR #1567, opened 2023-02-01
k8sutils: Update helm template to use apiVersion v2

mrueg:
See: https://helm.sh/docs/topics/charts/#the-apiversion-field


AhmedGrati (approved):
LGTM!

PR #1569, opened 2023-02-06
change master to main in scripts and docs

cdrage:
change master to main in scripts and docs

Changes the naming from master to main in the scripts and documentation.

Do not merge unless CI passes!

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,9 +2,9 @@ name: Go
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 env:
   # Avoid noisy outputs like "tput: No value for $TERM and no -T specified"
   TERM: dumb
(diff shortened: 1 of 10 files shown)


cdrage:
@AhmedGrati I am also disabling 1.18 since staticcheck no longer works with 1.18 with the GitHub actions: https://github.com/dominikh/staticcheck-action

I've also set a specific version for statticcheck to use 2023.1 due to the issues we keep having with CI regarding it.

cdrage:
> I think that issues still persist even if we disable staticcheck

Oooff, see any way to fix it?

AhmedGrati:
In the `Makefile` -> `validate`, if you remove staticcheck I think it works. Can you try it?

cdrage:
> In the `Makefile` -> `validate`, if you remove staticcheck I think it works. Can you try it?

Yeah, statticcheck IMO is causing too many problems at the moment. So I'll go ahead and change it.

AhmedGrati:
@cdrage In our code, we have two packages that require Go with version 1.19, but our CI runs on bot 1.18 and 1.19, so it fails on 1.18 version.

cdrage:
@AhmedGrati with this PR I'm also going to fix CI and just rip statticcheck out. Looks like CI is passing now!

AhmedGrati (approved):
LGTM

PR #1575, opened 2023-02-08
docs: update installation doc

cdrage:
docs: update installation doc

Some of the installation methods are either unmaintained / not up to
date anymore.

For example, we no longer package `.deb` packages.

The person using Kompose should be installing the binary directly
from GitHub rather than relying on their package manager.


AhmedGrati (approved):
LGTM! So much work on the documentation🤩.

PR #1613, opened 2023-03-12
Feat support security fsgroup

### LINKED ISSUE #1604: Add support `securityContext:fsGroup` (some container writes to volume using custom user/group id)
### Expected Behavior

Being able to convert a simple docker-compose with just one service -- pgadmin
and use the generated files directly

### Actual Behavior

As the volume is mounted on the containers owned by root, the container cannot write into it. pgadmin gives out error message **Permission denied: ‘/var/lib/pgadmin/sessions’**

One way to solve it is to add securityContext: fsGroup

```
      containers:
          ....
          name: pgadmin
          resources: {}
          volumeMounts:
            - mountPath: /var/lib/pgadmin
              name: pgadmin-data
          securityContext: # manutally added line
            fsGroup: 5050:5050 # manutally added line
``` 

In this way the pod will run without error.  

### Steps To Reproduce

kompose convert -f docker-compose.yaml --provider openshift

### Kompose Version

```Text
1.28.0 (c4137012e)
```

### Docker-Compose file

```YAML
version: '3.8'
volumes:
  pgadmin-data:

services:
  pgadmin:
    image: dpage/pgadmin4
    environment:
      PGADMIN_DEFAULT_EMAIL: dumb_pgadmin_user@email.com
      PGADMIN_DEFAULT_PASSWORD: pgadmin_password
    volumes:
      - pgadmin-data:/var/lib/pgadmin
```
```


### Anything else?

I haven't figure out why the same error doesn't occur when using docker-compose. Apparently it does not always mount with root:root.   `fsGroup` is just one way.  Another way I know would be using `initContainer` and chmod the path but I guess that would be more messy in terms of kompose conversion.  If we can just add some kompose label to indicate the fsGroup, then I think it would help a lot.  pgadmin is just one example.  grafana, and probably other commom images as well, all have this problem.

AhmedGrati:
#### What type of PR is this?

/kind feature

Add `security-group.fsGroup` for a specific pod, using a kompose label.

Fixes #1604.


cdrage (approved):
Awesome contribution :) thank you! LGTM!

/lgtm

PR #1656, opened 2023-07-02
Add additional Windows install method

### LINKED ISSUE #1655: Docs: Document the installation via winget on Windows

An alternative way to install kompose on windows

### Why is this needed?

winget is semi-official package manager (in that MS work on the same), and there is already a manifest for kompose under id `Kubernetes.kompose` which installs the executable from the Github releases.

Avinash-Bhat:
#### What type of PR is this?

/kind documentation

This PR supports additional  way to get Kompose installed in Windows using winget. 

Fixes #1655


kundan2707 (commented):
/lgtm

PR #1667, opened 2023-07-13
Feat add ns generation

### LINKED ISSUE #1657: Specify a namespace with convert?

Hello,

I'm writing up some documentation on how to use Konvert with the OpenShift Developer Sandbox. However, when I convert my docker-compose file then run `oc apply -f`, I get a bunch of errors around the Deployment Config and a missing namespace.

Is there anyway to fix this during the `convert` process?

```
oc apply -f ocp/
persistentvolumeclaim/databasevolume created
persistentvolumeclaim/logvolume01 created
service/openemr configured
persistentvolumeclaim/sitevolume created
resource mapping not found for name: "mysql" namespace: "" from "ocp/mysql-deploymentconfig.yaml": no matches for kind "DeploymentConfig" in version "v1"
ensure CRDs are installed first
resource mapping not found for name: "mysql" namespace: "" from "ocp/mysql-imagestream.yaml": no matches for kind "ImageStream" in version "v1"
ensure CRDs are installed first
resource mapping not found for name: "openemr" namespace: "" from "ocp/openemr-deploymentconfig.yaml": no matches for kind "DeploymentConfig" in version "v1"
ensure CRDs are installed first
resource mapping not found for name: "openemr" namespace: "" from "ocp/openemr-imagestream.yaml": no matches for kind "ImageStream" in version "v1"
ensure CRDs are installed first
```

### Why is this needed?

We want to make this as easy possible for those convert docker-compose files to a Kubernetes namespace/OpenShift project.

AhmedGrati:
#### What type of PR is this?

/kind feature

Implement namespace generation through the CLI flag `--namespace` or `-n`.

Fixes #1657.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -47,6 +47,7 @@ var (
 	ConvertReplicas              int
 	ConvertController            string
 	ConvertPushImage             bool
+	ConvertNamespace             string
 	ConvertPushImageRegistry     string
 	ConvertOpt                   kobject.ConvertOptions
 	ConvertYAMLIndent            int
@@ -122,6 +123,7 @@ var convertCmd = &cobra.Command{
 			GenerateNetworkPolicies:     GenerateNetworkPolicies,
 			BuildCommand:                BuildCommand,
 			PushCommand:                 PushCommand,
+			Namespace:                   ConvertNamespace,
 		}
 
 		if ServiceGroupMode == "" && MultipleContainerMode {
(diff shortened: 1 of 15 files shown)


cdrage:
Works for me! Thank you :)

cdrage (approved):

MikeTangoEcho:
@AhmedGrati @cdrage Is there a way to disable this feature ? Or when there is no namespace specified it doesn't add all resources to `default` namespace. I use kompose with helm chart generation, and the namespace is decided on helm release installation.

The lines to be modified would be.

https://github.com/AhmedGrati/kompose/blob/608f87669f63a93c7af5700ee3d17bdbbf197256/pkg/transformer/utils.go#L465-L468

I can make a PR if needed.

github-belhorma-bendebiche:
This is a breaking change. We currently create a namespace with other resources in it before installing the chart generated by kompose (and overriding the namespace using helm's flags). This flow is no longer possible: if we don't set a namespace, everything goes into `default`. If we do set a namespace, the chart is no longer installable (since the namespace already exists and we don't want the chart to create it).

cdrage:
> This is a breaking change. We currently create a namespace with other resources in it before installing the chart generated by kompose (and overriding the namespace using helm's flags). This flow is no longer possible: if we don't set a namespace, everything goes into `default`. If we do set a namespace, the chart is no longer installable (since the namespace already exists and we don't want the chart to create it).

On it! This shouldn't of been added, sorry about that. I totally missed it.

cdrage:
@github-belhorma-bendebiche Address in https://github.com/kubernetes/kompose/pull/1731

github-belhorma-bendebiche:
Thank you for the quick fix!

PR #1669, opened 2023-07-14
fix: remove lower case of the host generation

### LINKED ISSUE #1663: [BUG] kompose.service.expose label converts Helm variables to lowercase on translation into ingress host
### Expected Behavior

Having set
```
...
    labels:
      kompose.service.expose: "{{ .Release.Name }}.{{ .Release.Namespace }}.tld"
      ...
```
`kompose convert -c` should generate following ingress.yaml:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
...
spec:
  rules:
    - host: '{{ .Release.Name }}.{{ .Release.Namespace }}.tld'
    ...
```
Ingress annotation, however, translates perfectly fine.

### Actual Behavior

Instead generated ingress.yaml looks like this:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
...
spec:
  rules:
    - host: '{{ .release.name }}.{{ .release.namespace }}.tld'
    ...
```
and thus does not substitute corresponding Helm variables

### Steps To Reproduce

_No response_

### Kompose Version

```Text
1.2{8,9}.0
```

### Docker-Compose file

_No response_

### Anything else?

If this is intended behaviour and not a bug, i'd suspect a fair warning in Docs like "Do not attempt to use Helm variables inside labels / this particular label"

Thanks in advance!

AhmedGrati:
#### What type of PR is this?

/kind bug

From my POV, I don't see any need to make the host of the ingress in lowercase. If a user intended to make the host uppercase, we should not change it.

Fixes #1663.

### DIFF
--- a/pkg/loader/compose/compose.go
+++ b/pkg/loader/compose/compose.go
@@ -708,7 +708,7 @@ func parseKomposeLabels(labels map[string]string, serviceConfig *kobject.Service
 		case LabelSecurityContextFsGroup:
 			serviceConfig.FsGroup = cast.ToInt64(value)
 		case LabelServiceExpose:
-			serviceConfig.ExposeService = strings.Trim(strings.ToLower(value), " ,")
+			serviceConfig.ExposeService = strings.Trim(value, " ,")
 		case LabelNodePortPort:
 			serviceConfig.NodePortPort = cast.ToInt32(value)
 		case LabelServiceExposeTLSSecret:


kundan2707 (commented):
/lgtm

cdrage:
@kundan2707 Please only let code owners / maintainers use the LGTM.  As it automatically merges the PR.

kundan2707:
@cdrage 
I will take care of it onwards.

PR #1684, opened 2023-08-04
fix issue https://github.com/kubernetes/kompose/issues/1683

### LINKED ISSUE #1683: [BUG] pod env_file does not create configmaps
### Expected Behavior

docker-compose:
```yaml
version: "3"

services:
  pod:
    image: busybox
    env_file:
      - "envs/pod.env"
    restart: "no"

  deployment:
    image: busybox
    env_file:
      - "envs/deployment.env"
    restart: "always"
```
* Should create 2 **ConfigMaps**: envs-deployment-env & envs-pod-env. 

### Actual Behavior

Creates only 1 ConfigMap: envs-deployment-env 
* in Windows it fails to create any ConfigMap.
  **reason**:  service.EnvFile is not relative and it cause issues with some functions.

### Steps To Reproduce

2. Create envs directory
3. Create 2 files envs/deployment.env and envs/pod.env with the contents of `TEST=TEST`
4. Run command: `kompose convert`

### Kompose Version

```Text
1.30.0 
Branch: main
Commit: 17fbe3b4632cca0f61a7fda5f3b91034cdf5c5e4
```

### Docker-Compose file

```YAML
version: "3"

services:
  pod:
    image: busybox
    env_file:
      - "envs/pod.env"
    restart: "no"

  deployment:
    image: busybox
    env_file:
      - "envs/deployment.env"
    restart: "always"
```

### Anything else?

_No response_

realgam3:
#### What type of PR is this?
/kind bug

Fixes #1683

* Convert absolute paths to relative paths
* Replace Windows paths `\` to Linux paths `/`

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -226,9 +226,9 @@ func Convert(opt kobject.ConvertOptions) ([]runtime.Object, error) {
 	komposeObject.Namespace = opt.Namespace
 
 	// Get the directory of the compose file
-	composeFileDir, err := transformer.GetComposeFileDir(opt.InputFiles)
+	workDir, err := transformer.GetComposeFileDir(opt.InputFiles)
 	if err != nil {
-		log.Fatalf(err.Error())
+		log.Fatalf("Unable to get compose file directory: %s", err)
 	}
 
 	// convert env_file from absolute to relative path
@@ -241,7 +241,7 @@ func Convert(opt kobject.ConvertOptions) ([]runtime.Object, error) {
 				continue
 			}
 
-			relPath, err := filepath.Rel(composeFileDir, envFile)
+			relPath, err := filepath.Rel(workDir, envFile)
 			if err != nil {
 				log.Fatalf(err.Error())
 			}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -225,9 +225,10 @@ func Convert(opt kobject.ConvertOptions) ([]runtime.Object, error) {
 
 	komposeObject.Namespace = opt.Namespace
 
-	// Check if input files are specified
-	if len(opt.InputFiles) <= 0 {
-		log.Fatal("No input files specified")
+	// Get the directory of the compose file
+	composeFileDir, err := transformer.GetComposeFileDir(opt.InputFiles)
+	if err != nil {
+		log.Fatalf(err.Error())
 	}
 
 	// convert env_file from absolute to relative path
@@ -240,12 +241,7 @@ func Convert(opt kobject.ConvertOptions) ([]runtime.Object, error) {
 				continue
 			}
 
-			workDirAbs, err := filepath.Abs(filepath.Dir(opt.InputFiles[0]))
-			if err != nil {
-				log.Fatalf(err.Error())
-			}
-
-			relPath, err := filepath.Rel(workDirAbs, envFile)
+			relPath, err := filepath.Rel(composeFileDir, envFile)
 			if err != nil {
 				log.Fatalf(err.Error())
 			}
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -225,6 +225,11 @@ func Convert(opt kobject.ConvertOptions) ([]runtime.Object, error) {
 
 	komposeObject.Namespace = opt.Namespace
 
+	// Check if input files are specified
+	if len(opt.InputFiles) <= 0 {
+		log.Fatal("No input files specified")
+	}
+
 	// convert env_file from absolute to relative path
 	for _, service := range komposeObject.ServiceConfigs {
 		if len(service.EnvFile) <= 0 {
@@ -245,7 +250,7 @@ func Convert(opt kobject.ConvertOptions) ([]runtime.Object, error) {
 				log.Fatalf(err.Error())
 			}
 
-			service.EnvFile[i] = strings.Replace(relPath, "\\", "/", -1)
+			service.EnvFile[i] = filepath.ToSlash(relPath)
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/app/app.go
```diff
@@ -221,6 +222,30 @@ func Convert(opt kobject.ConvertOptions) {
...
+			if err != nil {
+				log.Fatalf(err.Error())
+			}
+
+			relPath, err := filepath.Rel(workDirAbs, envFile)
+			if err != nil {
+				log.Fatalf(err.Error())
+			}
+
+			service.EnvFile[i] = strings.Replace(relPath, "\\", "/", -1)
```
cdrage, line 245: We should be using https://pkg.go.dev/path/filepath#FromSlash not replacing `\\`.

Using fromSlash returns OS-specific path names.

cdrage (changes requested):

pkg/app/app.go
AhmedGrati, line 245: Can you please clarify more why we need this, because what you added in  the`kubernetes.go` file should solve the problem.

AhmedGrati (commented):
Thanks for the PR. left some comments.

pkg/app/app.go
realgam3: Sure,

**1)** To address the issue of excessively long names, I transitioned from absolute paths to relative ones.

**2)** A substitution occurs within the environment name. Hence, even when using relative paths, this substitution remains necessary. 
- Refer to the code [k8sutils.go#L855](https://github.com/kubernetes/kompose/blob/b83d4d4094ff887c087aa48f8dbab9803afe764c/pkg/transformer/kubernetes/k8sutils.go#L855). 
- It replaces "/" with "-". 
- Since Windows paths utilize "\\", this won't resolve the environment name issue.

realgam3: It doesn't solve the issue...
I need unix path and it still returns unix path

example:
```golang
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
	p := filepath.FromSlash("C:\\test\\main.go")
	fmt.Println(p)
}
```
output:
```text
C:\test\main.go
```

AhmedGrati (commented):
Thanks for the PR @realgam3. It looks good overall, but I think adding unit tests showing how the mentioned issues were fixed should be a great addition. wdyt?

AhmedGrati:
@realgam3 yes please you can add/update existing tests for pod envfile.

realgam3:
> @realgam3 yes please you can add/update existing tests for pod envfile.

I will, Thank you!

AhmedGrati (commented):
@realgam3 can we change the test `output-k8s-template.json` to yaml format?

domehahn:
Hi, it would be really cool if the commit was merged. Thanks.

AhmedGrati:
Hi @domehahn @realgam3, I'm getting time off, I'll review it next week. Thanks for your understanding!

pkg/transformer/kubernetes/kubernetes.go
```diff
@@ -220,7 +220,8 @@ func (k *Kubernetes) InitSvc(name string, service kobject.ServiceConfig) *api.Se
 
 // InitConfigMapForEnv initializes a ConfigMap object
 func (k *Kubernetes) InitConfigMapForEnv(name string, opt kobject.ConvertOptions, envFile string) *api.ConfigMap {
-	envs, err := GetEnvsFromFile(envFile)
+	workDir := filepath.Dir(opt.InputFiles[0])
```
cdrage, line 223: Using opt.InputFiles[0] throughout the code is kinda scary, I'd advise that we check to see that it's not an empty array.

What would be better is putting this into a function instead similar to https://github.com/kubernetes/kompose/blob/ddd5f337452952c355de5ec5d383b8beaff4c141/pkg/transformer/utils.go#L331

cdrage (changes requested):
Only one comment with regards to using `[0]` which isn't a good idea.

We also need tests added to kubernetes.go

Sorry for the delayed response in review! It's been a busy couple of weeks.

pkg/transformer/kubernetes/kubernetes.go
cdrage: Just checking to see if opt.InputFiles is empty and then fail is enough. 

It's just odd having it check the first element of the array.

AhmedGrati:
ping @realgam3. Any updates on this?

realgam3:
> ping @realgam3. Any updates on this?

Done

pkg/transformer/kubernetes/kubernetes.go
realgam3: It's now ready

realgam3:
@cdrage, @AhmedGrati anything else you need from me?

AhmedGrati:
@realgam3 gonne review it ASAP.

realgam3:
> @realgam3 gonna review it ASAP.

Any updates / comments?
I don't want to miss the opportunity to push this update, this is an important fix for windows and linux users.

script/test/fixtures/configmap-pod/output-os.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - name: BAR
+          valueFrom:
+            configMapKeyRef:
+              key: BAR
+              name: bar-env
+        - name: FOO
+          valueFrom:
+            configMapKeyRef:
+              key: FOO
+              name: bar-env
```
AhmedGrati, line 44: @realgam3 I see here that you are referencing a config map which is not generated. I think that we should handle it in the `openshift.go` file.

realgam3: @AhmedGrati Great catch! I'm not sure how this one slipped past me. :S
Fixed and pushed....

AhmedGrati (approved):
Leaving final approve to @cdrage!

cdrage (approved):

PR #1709, opened 2023-09-17
feat: adding support for --profile cmd option

axel7083:
#### What type of PR is this?

/kind feature

When using docker compose we can use the [profiles](https://docs.docker.com/compose/compose-file/05-services/#profiles) attribute. 

> profiles defines a list of named profiles for the service to be enabled under. If unassigned, the service is always started but if assigned, it is only started if the profile is activated.

In the current behaviour of kompose, if the compose file is using profiles, it will not process the services. This PR allows to add the `--profile` cmd command (repeatable) to send a list of string to the loader, that will be able to process the file and enable the services.

Fixes https://github.com/kubernetes/kompose/issues/1708 https://github.com/kubernetes/kompose/issues/1458 https://github.com/kubernetes/kompose/issues/1448

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -201,7 +201,7 @@ func init() {
 
 	convertCmd.Flags().IntVar(&ConvertYAMLIndent, "indent", 2, "Spaces length to indent generated yaml files")
 
-	convertCmd.Flags().StringArrayVar(&ConvertProfiles, "profile", []string{}, `Specify the profile to use, can use multiple.`)
+	convertCmd.Flags().StringArrayVar(&ConvertProfiles, "profile", []string{}, `Specify the profile to use, can use multiple profiles`)
 
 	// In order to 'separate' both OpenShift and Kubernetes only flags. A custom help page is created
 	customHelp := `Usage:{{if .Runnable}}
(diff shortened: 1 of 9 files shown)


cdrage:
The code LGTM and implementation! But you'll have to add a test if that's okay for it to be all merged in:)

axel7083:
> The code LGTM and implementation! But you'll have to add a test if that's okay for it to be all merged in :)

I've added a test in the client. To do so, I added an example of docker-compose with some profile, and using the tool, I ensure it matches the expected output.

cmd/convert.go
```diff
@@ -199,6 +201,8 @@ func init() {
 
 	convertCmd.Flags().IntVar(&ConvertYAMLIndent, "indent", 2, "Spaces length to indent generated yaml files")
 
+	convertCmd.Flags().StringArrayVar(&ConvertProfiles, "profile", []string{}, `Specify the profile to use, can use multiple.`)
```
cdrage, line 204:
```suggestion
	convertCmd.Flags().StringArrayVar(&ConvertProfiles, "profile", []string{}, `Specify the profile to use, can use multiple profiles`)
```

cdrage (changes requested):
One TINNYYYY change (no period at end of sentence and little bit of grammar change), but otherwise LGTM!

cdrage (approved):

PR #1727, opened 2023-10-07
[Documentation] Add Linux install command for Mint

merobi-hub:
#### What type of PR is this?

/kind documentation

The Linux installation command does not work for Linux Mint: the placement of the cURL output parameter and argument (`-o kompose`) causes the command not to execute. This adds a command specifically for Linux Mint. Tested on Mint and Ubuntu.


cdrage (commented):
Sorry, we wont be adding instructions for Mint :( Using -o before not after is due to an old `curl` version.

PR #1739, opened 2023-10-17
remove status field from generated yaml

shubhindia:
/kind feature
#### What type of PR is this?
Whenever the k8s yamls are generated, since status field is part of APIs, it is being added in the generated yamls, which is fine but should not be there since the objects are not applied yet.

This just adds a simple hack to remove the status field from generated yamls

Fixes https://github.com/kubernetes/kompose/issues/975

### DIFF
--- a/pkg/transformer/utils.go
+++ b/pkg/transformer/utils.go
@@ -275,6 +275,9 @@ func ConfigAnnotations(service kobject.ServiceConfig) map[string]string {
 // Print either prints to stdout or to file/s
 func Print(name, path string, trailing string, data []byte, toStdout, generateJSON bool, f *os.File, provider string) (string, error) {
 	file := ""
+	// simple hack to remove status from the output
+	re := regexp.MustCompile(`(?s)status:.*`)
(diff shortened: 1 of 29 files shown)


AhmedGrati (commented):
Thanks for the PR @shubhindia. I think removing the status should be okay, as it is not really needed. However, You should change the manifests in the e2e tests accordingly in order for the e2e tests to pass successfully. wdyt @cdrage?

cdrage:
> Thanks for the PR @shubhindia. I think removing the status should be okay, as it is not really needed. However, You should change the manifests in the e2e tests accordingly in order for the e2e tests to pass successfully. wdyt @cdrage?

Yup! Fix the tests please and we can merge this in. I agree that we do not need status in the output.

AhmedGrati:
ping @shubhindia

shubhindia:
Acknowledged @AhmedGrati . Will push the changes by Monday.

shubhindia:
closed in favour of https://github.com/kubernetes/kompose/pull/1746

PR #1743, opened 2023-10-20
fix: env_file configmap name length

### LINKED ISSUE #1738: [BUG] Generated ConfigMap name too long
### Expected Behavior

A kubernetes resource name has a limit of 63 chars. When converting a docker-compose file with a `env_file` attribute it uses the full path of the resource as the name. This quickly exceeds the 63 char limit.

### Actual Behavior

It should restrict the name to filename or relative path (relative to the compose file) and stay withing 63 chars.

### Steps To Reproduce

have a docker-compose with a service.
Add to this service a `env_file` attribute.
convert with kompose
Result is a ConfigMap where the name is a reflection of the full file path. 

If the configmap name exceeds 63 then the resources cannot be deployed.

### Kompose Version

```Text
1.31.2
```

### Docker-Compose file

_No response_

### Anything else?

_No response_

ThijsBroersen:
fix filename to configmap name transformer

#### What type of PR is this?

bug

Fixes #1738

~~I was unable to run tests, could not figure out what system requirements were needed to run the tests.~~
Tests are adjusted to match new configmap name


pkg/transformer/kubernetes/k8sutils.go
```diff
@@ -851,8 +851,15 @@ func GetContentFromFile(file string) (string, error) {
 // FormatEnvName format env name
 func FormatEnvName(name string) string {
 	envName := strings.Trim(name, "./")
+	// only take string after the last slash only if the string contains a slash
+	if strings.Contains(envName, "/") {
```
cdrage, line 855: Thank you! This helps, does this break anything though @AhmedGrati with regards to how we use env?

This also cleans up the really long configmap names, but I'm unsure if they would have conflicts when there are multiple containers.

If you can also add a test case for this under k8sutils_test.go that'd be awesome. Otherwise, good PR 👍

cdrage (changes requested):

pkg/transformer/kubernetes/k8sutils.go
ThijsBroersen: Thanks!
I think it will work as long as env_file names are unique. When different env_file's with the same name are references (from different directories) it will clash. 
Could also just only cap at the last, let's say 60 chars? And not filter by last `/` ?

ThijsBroersen: Do you want any changes? If not could it be merged soon? I could really use this in a new version of Kompose.

cdrage: You're right, I think that the names generated should be from env_file names from the last slash.. I really doubt that there will be multiple env file names in different directories for a docker compose project.

HOWEVER if there is, I think that they should be the same name, but one of them `filename_1`, etc. if there are more of the same file AND Kompose does a warning.

However, I don't feel comfortable merging the PR until we have these two tasks:
- [ ] tests added to k8sutils_test.go
- [ ] test for if there are two env file names in different directories
- [ ] implement renaming to filename_number if there are multiple env files of the same name but in different directories.

ThijsBroersen: Ok, just took a short browse. Added a new `env-multiple` fixture and test but now sure how to implement this. My thinking would be to check existing configmaps before inserting the new. If a configmap with the same name exists, append a 1 and try again. If that also exists a 2, and so on..
Could you give me a pointer? Or perhaps someone else wants to do this.

.gitpod.yml
```diff
@@ -0,0 +1,11 @@
...
+# Please adjust to your needs (see https://www.gitpod.io/docs/introduction/learn-gitpod/gitpod-yaml)
+# and commit this file to your remote git repository to share the goodness with others.
+
+# Learn more from ready-to-use templates: https://www.gitpod.io/docs/introduction/getting-started/quickstart
+
+tasks:
+  - init: go get && go build ./... && go test ./... && make
+    command: go run .
+
+
```
AhmedGrati, line 11: I think that this should not be added.

pkg/transformer/kubernetes/k8sutils.go
AhmedGrati: Sorry, I didn't actually get it @ThijsBroersen. Can you please clarify it more?

AhmedGrati (commented):
Thanks for the PR @ThijsBroersen. Maybe a unit test for this would make it even better 🙏.

cdrage:
I'm good with this PR! LGTM! I understand about writing a golang test, I've opened up an issue here so we make sure we add a future test: https://github.com/kubernetes/kompose/issues/1745

cdrage (approved):

cdrage:
You'll have to rebase the issue to pass the new tests

ThijsBroersen:
I rebased as requested but the new tests will likely still fail because I did not found a way yet to take into account duplicate names. Should I comment out the new test in `new_tests.sh` ?

AhmedGrati:
Thanks for the contribution @ThijsBroersen. I'm Gonna take care of the tests.

PR #1744, opened 2023-10-24
fix: make transforming more deterministic

AhmedGrati:
#### What type of PR is this?

/kind bug

This PR makes the transformation deterministic by eliminating looping over maps. Instead, we are now looping over sorted arrays of keys.


cdrage:
This is amazing 💯 💯 thank you!

Issue #1773, opened 2023-12-04
[BUG]  Updated doc installation.md


utkarsh-singh1:
### Expected Behavior

As fedora 24-26 has already reached end of life (fedora 26 reached at 05-2018), remove fedora installation instruction from the Installation docs. Also,

-  Kompose latest release (1.32) is ahead of release candidate on Fedora. 
-  Latest release can be easily installed using scripts. 

### Actual Behavior

Remove current fedora installation instruction.

### Kompose Version

```Text
1.31.2
```

PR #1774, opened 2023-12-04
Removed fedora installation instruction from installation.md

utkarsh-singh1:
#### What type of PR is this?
/kind documentation

Installation instruction for linux, it can Also be installed with brew or using go install method.

Fixes # 1773


utkarsh-singh1:
Thanks, @cdrage, for approval.

Fixes #1773

PR #1775, opened 2023-12-07
feat(1765): remove useless keys from the output yml

### LINKED ISSUE #1758: Remove other useless keys in output

```yaml
▶ cat frontend-deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    kompose.cmd: kompose convert
    kompose.version: 1.31.1 (37a1dc26a)
  creationTimestamp: null
  labels:
    io.kompose.service: frontend
  name: frontend
spec:
  replicas: 1
  selector:
    matchLabels:
      io.kompose.service: frontend
  strategy: {}
  template:
    metadata:
      annotations:
        kompose.cmd: kompose convert
        kompose.version: 1.31.1 (37a1dc26a)
      creationTimestamp: null
      labels:
        io.kompose.network/downloads-default: "true"
        io.kompose.service: frontend
    spec:
      containers:
        - image: cdrage/frontend
          name: frontend
          ports:
            - containerPort: 8080
              hostPort: 8080
              protocol: TCP
          resources: {}
      restartPolicy: Always
status: {}
```

### Why is this needed?

There are keys which are useless in the output we should remove such as:

* strategy {}
* resources: {}
* status: {}

martinjirku:
#### What type of PR is this?
/kind feature

This PR transforms yaml output by removing map[string]interfaces{} with no keys, or the nil fields. This PR utilizes the current implementation of the yaml transformation process in the "kompose" `(1) object ->  (2) []byte (json) -> (3) interface{} -> (4) []byte (yaml)`. Between steps 3 and 4, there is a chance to clean the data structure by removing any empty fields.

Fixes #1758

IMHO, The empty artifacts  (such as strategy {} , resources: {},...) are present because we have e.g.  in the [`DeploymentSpec.Strategy`](https://github.com/kubernetes/kubernetes/blob/89dfbebe2e823bd0b13e4b65b8460a058c7258a3/pkg/apis/apps/types.go#L385) struct not the reference (a pointer) to the struct, since this change would be outside of the kompose scope and I am not sure if it is feasible, I've decided to fix it by removing all empty fields.

Is this correct approach I should proceed with, or should I open (or search if a ticket exists) in the k8s/kubernetes repo for change in the API app types?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,5 +77,8 @@ tags
 
 .DS_Store
 
+# VSCode specific
+.vscode
+
 # Client Test generated files
 client/testdata/generated
--- a/pkg/transformer/kubernetes/k8sutils.go
+++ b/pkg/transformer/kubernetes/k8sutils.go
@@ -280,6 +280,34 @@ func marshal(obj runtime.Object, jsonFormat bool, indent int) (data []byte, err 
 	return
 }
 
+// remove empty map[string]interface{} strings from the object
+//
+// Note: this function uses recursion, use it only objects created by the unmarshalled json.
+// Passing cyclic structures to removeEmptyInterfaces will result in a stack overflow.
+func removeEmptyInterfaces(obj interface{}) interface{} {
+	switch v := obj.(type) {
+	case []interface{}:
+		for i, val := range v {
+			if valMap, ok := val.(map[string]interface{}); (ok && len(valMap) == 0) || val == nil {
+				v = append(v[:i], v[i+1:]...)
+			} else {
+				v[i] = removeEmptyInterfaces(val)
+			}
+		}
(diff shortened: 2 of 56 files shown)


martinjirku:
I did not update docker tests, I will update it in next commit

AhmedGrati:
@martinjirku any updates on this?

martinjirku:
@AhmedGrati I've got much at work before Christmas. Still working... I've updated cca 40 tests sofar.

cdrage:
AMAZING work! Thank you so much 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 

LGTM!

PR #1803, opened 2024-01-15
Add new examples to kompose

cdrage:
Add new examples to kompose

#### What type of PR is this?

/kind cleanup

Fixes the current broken examples by:

* Removing all the old incompatible ones (we do not really support v3
  anymore or v2... since switching libraries)
* Uses quay.io/kompose/web as our front end example which is a fork of
  the guestbook-go kubernetes examples

Closes https://github.com/kubernetes/kompose/issues/1757

Test using docker-compose (you'll see it come up!), then try with
kompose :)


AhmedGrati (approved):
Looks good overall, thanks @cdrage we really need this 🎉🙏! I left a small comment.

Issue #1818, opened 2024-02-04
[BUG] Kompose Client can't create pods / cronjobs


realgam3:
### Expected Behavior

1) Default kubernetesController should be empty string ("")
2) kubernetesController should validate empty kubernetesController  as valid (just like the kompose cli)

### Actual Behavior

The validation of kubernetesController (source code):
```golang
kubernetesController := kubernetesProvider.Controller
if *kubernetesController != string(DEPLOYMENT) && *kubernetesController != string(DAEMONSET) && *kubernetesController != string(REPLICATION_CONTROLLER) {
	return fmt.Errorf(
		"unexpected Value for Kubernetes Controller field. Possible values are: %v, %v, and %v", string(DEPLOYMENT), string(DAEMONSET), string(REPLICATION_CONTROLLER),
	)
}
```

The validation of pod/cronjob (source code):
```golang
func (opt *ConvertOptions) IsPodController() bool {
        // opt.IsDeploymentFlag = true, opt.Controller = "deployment"
	return opt.IsDeploymentFlag || opt.IsDaemonSetFlag || opt.IsReplicationControllerFlag || opt.Controller != ""
}
```

### Steps To Reproduce

```golang
package main

import (
	"fmt"
	"github.com/kubernetes/kompose/client"
	"log"
	"os"
)

func main() {
	cli, err := client.NewClient(client.WithSuppressWarnings())
	if err != nil {
		log.Fatal("Could not create client")
	}

	annotations := false
	controller := ""
	res, err := cli.Convert(client.ConvertOptions{
		InputFiles: []string{"./docker-compose.yaml"},
		OutFile:    os.DevNull,
		//ToStdout:               false,
		ToStdout:               true,
		WithKomposeAnnotations: &annotations,
		Provider: client.Kubernetes{
			Chart:      false,
			Controller: &controller,
		},
	})
	if err != nil {
		log.Fatal("Could not convert")
	}
	fmt.Println(res[0])
}

```

### Kompose Version

```Text
1.32.0
```

### Docker-Compose file

```YAML
version: '3'

services:
  challenge:
    image: busybox:stable-glibc
    labels:
      kompose.cronjob.schedule: "* * * * *"
      kompose.cronjob.concurrency_policy: "Forbid"
      kompose.cronjob.backoff_limit: "0"
    command:
      - "bash"
      - "-c"
      - "echo hello from cron"
    restart: "no"
```

PR #1819, opened 2024-02-04
fix client/convert (issue https://github.com/kubernetes/kompose/issue…

realgam3:
#### What type of PR is this?
/kind bug

* Client Default Controller should be empty string ("") and empty string should be a valid value.

Fixes [#1818](https://github.com/kubernetes/kompose/issues/1818)

@AhmedGrati I think you should review this since it's your feature.


cdrage (approved):

cdrage:
Thanks for the quick fix!

PR #1824, opened 2024-02-12
#1821 update Dockerfile to alwyas use head Alpine

### LINKED ISSUE #1821: [BUG] Bump Alpine
### Expected Behavior

Use supported version of Alpine 

### Actual Behavior

Currently built against 3.9

### Steps To Reproduce

_No response_

### Kompose Version

```Text
Na
```

### Docker-Compose file

_No response_

### Anything else?

I am simply an outside observer, admittedly I don't even use this software.
With that said your Dockerfile was linked to me as a meme as it is running a version of Alpine that has been out of support since 2021.
https://github.com/kubernetes/kompose/blob/main/Dockerfile#L7
Just a heads up, feel free to insta close this but you might want to think about a more scratch base or bumping Alpine or if this logic is not used in your workflow removed from the repo.

thelamer:
#### What type of PR is this?

/kind bug

This updates the current Dockerfile to support aarch64 and armv7 as well, it also bumps to always use head Alpine instead of the previous hard coded 3.9. Given your instructions only use busybox for the ability to cd into the /opt directory this can likely be based down to scratch with some minor doc updates, but that is not the purpose of this PR and the space savings are negligible to potentially break external workflows. 

Fixes #1821

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,17 @@
-FROM alpine:3.9 as builder
+# Alpine Builder
+FROM alpine as builder
 
 RUN apk add --no-cache curl
 COPY ./build/VERSION VERSION
-RUN version=$(cat VERSION) && curl -L "https://github.com/kubernetes/kompose/releases/download/v${version}/kompose-linux-amd64" -o kompose
+RUN \
+  version=$(cat VERSION) && \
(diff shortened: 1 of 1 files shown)


cdrage (approved):
Woo this is an awesome improvement! LGTM!

/approve

/lgtm

You'll just have to sign the CLI and this will be merged 🎉

PR #1826, opened 2024-02-14
fix issue 1778 update1

### LINKED ISSUE #1778: [BUG] Kompose does not mount secrets correctly
### Expected Behavior

secrets defined in compose.yaml should be mounted in the pods in `/run/secrets/<secret>`

### Actual Behavior

`/run/secrets/<secret>` is mounted as a directory

### Steps To Reproduce

1. Add secrets to compose.yaml

```
services:
  myservice:
    ...
    secrets:
      - config-ini

secrets:
  config-ini:
    file: secrets/Config.ini
```

2. kompose convert

3. examine generated pod config file:

The result is:
```
volumeMounts:
  - mountPath: /run/secrets/config-ini
     name: config-ini
```

Which appears correct at a glance, but when running, I get the error that `config-ini` is a directory.

### Kompose Version

```Text
1.31.2 (HEAD)
```

### Docker-Compose file

_No response_

### Anything else?

I was able to fix the pod file by adding subPath:
```
volumeMounts:
  - mountPath: /run/secrets/config-ini
    subPath: config-ini
    name: config-ini
```

With subPath, it gets mounted as a file instead of a directory.

I don't know k8 very well, I would guess it's mounted as a directory because secrets can hold more than one key-value pair. In that case it wouldn't necessarily make sense to put all the key-value pairs into one file, especially if each key should be its own file.

But with docker compose, secrets are only one key->value, so that behavior should be translated.

sosan:
in case /run/secrets/MARIADB_PASSWORD and secrets: MARIADB_PASSWORD are in uppercase, reformat the value of MARIADB_PASSWORD to mariadb-password. Also, reformat the secretconfig of the secrets. also getSecretPathsLegacy returns secretSubPath.

#### What type of PR is this?
/kind cleanup

updated fix to contemplate mays in secrets:
```
version: "3.9"
services:
  database:
    image: mariadb:11
    deploy:
      replicas: 1
    environment:
      MARIADB_ROOT_PASSWORD_FILE: /run/secrets/MARIADB_PASSWORD
    secrets:
      - MARIADB_PASSWORD
secrets:
  MARIADB_PASSWORD:
    file: "./mariadb-password.txt"
```

Fixes #1778


cdrage (approved):

microsec-ambrose:
This caused a regression in our Compose file:

```yml
services:
  command: ["cat", "/etc/hello_world.yaml"]
  secrets:
    - source: hello-world
      target: /etc/hello_world.yaml
secrets:
  hello-world:
    file: config/hello_world.yaml
```

I think only the `source` needs to be renamed, and the `target` left alone, unless the `source` is of a different name.

microsec-ambrose:
The output of the original Compose by @sosan should be:

```yml
    spec:
      containers:
        - env:
            - name: MARIADB_ROOT_PASSWORD_FILE
              value: /run/secrets/MARIADB_PASSWORD
          image: mariadb:11
          name: database
          volumeMounts:
            - mountPath: /run/secrets
              name: mariadb-password
              subPath: MARIADB_PASSWORD
      restartPolicy: Always
      volumes:
        - name: mariadb-password
          secret:
            items:
              - key: mariadb-password
                path: MARIADB_PASSWORD
            secretName: mariadb-password
```

cdrage:
> I would implement this solution differently, without touching any `environment`s:
> 
> 
> 
> 1. If `target` is specified, only ever transform the `source` field.
> 
> 2. If `target` is not specified, fill it in as `/run/secrets/{source_before_transform}`.
> 
> 
> 
> The output of the original Compose by @sosan should be:
> 
> 
> 
> ```yml
> 
>     spec:
> 
>       containers:
> 
>         - env:
> 
>             - name: MARIADB_ROOT_PASSWORD_FILE
> 
>               value: /run/secrets/MARIADB_PASSWORD
> 
>           image: mariadb:11
> 
>           name: database
> 
>           volumeMounts:
> 
>             - mountPath: /run/secrets
> 
>               name: mariadb-password
> 
>               subPath: MARIADB_PASSWORD
> 
>       restartPolicy: Always
> 
>       volumes:
> 
>         - name: mariadb-password
> 
>           secret:
> 
>             items:
> 
>               - key: mariadb-password
> 
>                 path: MARIADB_PASSWORD
> 
>             secretName: mariadb-password
> 
> ```

Hi! Could you open an issue so we make sure that we track this?

PR #1832, opened 2024-02-19
requests and limits by default as flag --requestslimits

sosan:
#### What type of PR is this?

/kind feature

The `getResourceRequests` and `getResourceLimits` functions now provide predefined CPU and memory requests, so you don't need to fill them manually.
We also added the -l or --limitsrequests flag to force the requests and limits to be generated.

### DIFF
--- a/cmd/convert.go
+++ b/cmd/convert.go
@@ -129,6 +129,7 @@ var convertCmd = &cobra.Command{
 			BuildCommand:                BuildCommand,
 			PushCommand:                 PushCommand,
 			Namespace:                   ConvertNamespace,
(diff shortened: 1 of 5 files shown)


cdrage:
> Thanks for the patch @sosan. I honestly don't see the purpose of adding a flag to add requests and limits, as these options can be specified directly from the compose spec. I do believe that this would add more maintenance efforts from our side, which can be avoided by simply letting users specify resources requirements in the spec file directly. Please let me know if I'm missing something! Thoughts on this @cdrage?

I have to agree as well, I don't see the purpose of adding this as another flag :( we are trying to reduce the amount of parameters, not increase it.

Thank you for your hard work though, but unfortunately I agree that we shouldn't add it.

PR #1835, opened 2024-02-29
chore: replace AhmedGrati with TessaIO

TessaIO:
The `AhmedGrati` account is locked and should be replaced with `TessaIO`. You can check that both have the same e-mail.


cdrage (approved):

cdrage:
Confirmed developer is the same as the old account @AhmedGrati .

LGTM

Issue #1635, opened 2023-04-26
[Feature] customizing service name


s33kers:
### What would you like to be added?

Now when creating Service and Deployment out of docker-compose.yaml service name is used for Kubernetes resource naming which cannot be customized out-of-the-box (e.g. providing environment variables for docker-compose.yaml during `docker compose config`)

- Option would be to have kompose label (e.g. kompose.name-override) which would allow modifying Kubernetes resource name
- Option would be to provide flag during `kompose convert` which would apply it as a prefix for all resources

Using `container_name` only changes container name within Pod.
Using `kompose.service.group` partially lets us achieve goal because Services are left with the same name as it is in docker-compose.yaml

Currently, we are solving that with string replace
docker-compose.yaml
```
services:
  prefix-service-a:
    image: xxx:latest
    ports:
      - "9999:9999"

  prefix-service-b:
    image: yyy:latest
    ports:
      - "8888:8888"
    environment:
      URL: http://prefix-xxx:9999
```
and then applying `sed -i -e "s/prefix-/${random-uuid}/g" docker-compose.yaml` before `kompose convert`

### Why is this needed?

Kubernetes Service name cannot be customized when generating resources from docker-compose.yaml

Tobias-Pe:
+1 This would also be great for Loadbalance Service types to remove the -tcp suffix.
Currently, I have to do this with a find and replace command or else my service endpoints have different names than my docker compose declares.

sosan:
> +1 This would also be great for Loadbalance Service types to remove the -tcp suffix. Currently, I have to do this with a find and replace command or else my service endpoints have different names than my docker compose declares.

removing the "-tcp"/"-udp" suffix by default, providing an option via a label, argument, ....?

sosan:
> +1 This would also be great for Loadbalance Service types to remove the -tcp suffix. Currently, I have to do this with a find and replace command or else my service endpoints have different names than my docker compose declares.

this labels for more granular control?
```
kompose.service.suffix-lb: "-new-suffix"
kompose.service.suffix-lb: ""
kompose.service.suffix-lb-tcp: "-tcp"
kompose.service.suffix-lb-udp: "-udp"
```

`kompose.service.suffix-lb: ""` will remove both suffix for tcp and udp

or something:

`kompose.service.no-suffix-lb: "yes"`

opinions?

cdrage:
[...]

PR #1837, opened 2024-03-10
customize resource name individually via labels kompose.service.name_override

sosan:
#### What type of PR is this?

/kind feature

this PR is the second part from #1833 and suggests labels to allow individual customization. Uses `kompose.service.name_override` label, but can change label name 

Feature #1635

tests `Test_parseKomposeLabels` encounters a corner case scenario where the resource name that being processed starts or ends with a hyphen '-' or any other character

// This is a corner case that is expected to fail because
// it does not account for scenarios where the string
// starts or ends with a '-' or any other character
// this test will fail with current tests


TessaIO:
Thanks @sosan, I would review this once the other PR is merged!

cdrage (commented):
/lgtm

cdrage (approved):

PR #1838, opened 2024-03-11
chore(deps): bump github.com/compose-spec/compose-go

microsec-ambrose:
#### What type of PR is this?

/kind feature

Upgrades `compose-go` to support new syntax.

Currently, any output from `docker compose config` including `env_files` is not accepted by Kompose. Docker Compose v2.24.7 changed the output schema.

N.A.


TessaIO (changes requested):
Thanks @microsec-ambrose for the changes. Some things are broken now. Do you have some time to commit to fix them?

microsec-ambrose:
compose-go only supports 1.21 and above due to the use of the `cmp` package (see https://github.com/compose-spec/compose-go/pull/531).

I have updated the PR to upgrade Go too.

pkg/loader/compose/compose.go
```diff
@@ -694,6 +695,12 @@ func parseEnvironment(composeServiceConfig *types.ServiceConfig, serviceConfig *
 	}
 }
 
+func parseEnvFiles(composeServiceConfig *types.ServiceConfig, serviceConfig *kobject.ServiceConfig) {
```
cdrage, line 698: Missing a test function for this new one, if you're able to add a test for it that'd be awesome

cdrage (changes requested):
Sorry for the delay! Been a busy past couple of weeks for me.

Awesome changes, just one comment about a test, but that's it.

TessaIO (approved):

PR #1850, opened 2024-03-31
added general suffix for LB overrider

sosan:
#### What type of PR is this?
/kind feature

Loadbalance Service customize the suffix with labels.
Empty value remove suffix

Fixes #https://github.com/kubernetes/kompose/issues/1635#issuecomment-2015707539


TessaIO:
Referring to https://github.com/kubernetes/kompose/issues/1635#issuecomment-2031874480 should we close this @sosan @cdrage ?

cdrage:
> Referring to [#1635 (comment)](https://github.com/kubernetes/kompose/issues/1635#issuecomment-2031874480) should we close this @sosan @cdrage ?

Yes, we should close this. Sorry about the work @sosan we should try to cut down on parameters / suffix / prefix control over smaller stuff. Big stuff like services, components, etc, yes. But if we give too much control over suffix and prefix for every part, we'd have to add them to everything.

PR #1879, opened 2024-05-29
feature: change example to use non-root

cdrage:
feature: change example to use non-root

#### What type of PR is this?

/kind feature

This changes our example to use non-root so it runs well on non-root
clusters (ex. openshift).

We also add debugging tools so we can safely explore the container.

/tmp is also added as the directory for the replica or else it fails.

N/A

### DIFF
--- a/examples/compose.yaml
+++ b/examples/compose.yaml
@@ -2,16 +2,16 @@ services:
 
   redis-leader:
     container_name: redis-leader
-    image: redis:latest
+    image: redis
     ports:
       - "6379"
 
   redis-replica:
     container_name: redis-replica
-    image: redis:latest
+    image: redis
     ports:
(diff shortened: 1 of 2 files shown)


examples/compose.yaml
```diff
@@ -2,19 +2,19 @@ services:
...
       - "6379"
-    command: redis-server --replicaof redis-leader 6379
+    command: redis-server --replicaof redis-leader 6379 --dir /tmp
 
   web:
     container_name: web
     image: quay.io/kompose/web
     ports:
-      - "8080:8080"
+      - "8080:8080"
```
TessaIO, line 20: Missing EOF

TessaIO (approved):
Thanks @cdrage!

PR #1884, opened 2024-05-29
cleanup: removes uneeded annotation when not using network policy

cdrage:
cleanup: removes uneeded annotation when not using network policy

#### What type of PR is this?

/kind cleanup

Removes the network policy annotation which is not needed (we are not
generating network policy)

Fixes https://github.com/kubernetes/kompose/issues/1759


cdrage:
Not sure what I'm doing wrong here @TessaIO :( it's been a while since I've tested this locally, but I cant seem to get it to work with test-cmd and such on mac M1. Are you able to fix these tests by any change?

cdrage:
Failing because https://github.com/kubernetes/kompose/pull/1886 needs to go in

TessaIO (approved):

PR #1886, opened 2024-05-30
tests: fix flakey tests

cdrage:
tests: fix flakey tests

#### What type of PR is this?

/kind cleanup

Fixes the two flakey tests. One which is returning them in random
orders..

The other which is not available to run on mac due to not being able
"access" the folders.

N/A

N/A


cdrage:
Going to merge this in so that we can get the other tests passing, lmk if there is issues here.

Since this is to fix CI and does not change any underlying code, I will merge.

PR #1915, opened 2024-07-31
[completion] Add header #compdef

sazriel26:
#### What type of PR is this?

/kind bug

Generation of zsh completion misses header `#compdef kompose` as updated in reference:
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/completion/completion.go#L174

N/A

Not familiar with Go, but looks confident about 🤗 

Since the completion is not "native" but converted from `bash`, the directive `compdef <function> <command>` looks not mandatory (this is the second line to append when completion is zsh-native)

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -85,6 +85,8 @@ in order to generate ZSH completion support.
 */
 func runCompletionZsh(out io.Writer, kompose *cobra.Command) error {
 	zshInitialization := `
+#compdef kompose
+
 __kompose_bash_source() {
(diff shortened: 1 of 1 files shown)


cdrage (approved):
This LGTM but unsure regarding why CI is failing, any idea @TessaIO ?

TessaIO:
@sazriel26 can you rebase please to make the tests pass

cdrage:
Sorry for the late merge, LGTM now! Tested it locally and it works.

PR #1944, opened 2024-10-27
bug: fix named volume detection logic

ms-jpq:
#### What type of PR is this?

/kind bug

Currently an input with a named volume defined in the same `docker-compose.yml` would produce an error like so:

> WARN File don't exist or failed to check if the directory is empty: stat :/app/data: no such file or directory

minimal example:

```yaml
---
services:
  example:
    image: bash
    volumes:
      - data:/app/data

volumes:
  data:
```

This bug was [introduced here](https://github.com/kubernetes/kompose/pull/1852/files#diff-1ff9d325f365a9e62567ac976258ba7a180503c6b53820c8f04b174d69b5e23fR960), the code for checking say `/app/data` in the example below, neglected to account for volumes like the one shown above.

```yaml
volumes:
  - /app/data
```

This PR simply adds a check to `data:`, as in `data:/app/data`, where as currently this is treated identically to just `/app/data`.

Fixes #


cdrage:
Failing tests :( are you able to check the failed `TestMultipleContainersInPod` test?

PR #1950, opened 2024-10-30
Fix issue 1942

hurzelpurzel:
#### What type of PR is this?
/kind bug

PVC names will now in the valid format

Fixes # 1942

go.mod was changed to point to a valid openshift api version

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,3 +82,6 @@ tags
 
 # Client Test generated files
 client/testdata/generated
+
+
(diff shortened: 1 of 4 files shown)


cdrage:
Thank you for the fix as well as the fix with compiling / mod go 🚀 🚀 🚀 🚀 🚀 🚀

cdrage:
This code LGTM!

PR #1958, opened 2024-11-07
improve error output of git commands

### LINKED ISSUE #1957: [BUG] kompose fails with getAbsBuildContext failed: exit status 128
### Expected Behavior

I have a gitlab pipeline that runs kompose. Recently I updated the docker image and started running into the following error from kompose:
```
$ kompose convert --provider=openshift --build build-config --build-repo "https://${CI_SERVER_HOST}/${CI_PROJECT_PATH}.git" --build-branch "$CI_COMMIT_REF_NAME" -f docker-compose.k8s-resolved.yml -o $KOMPOSE_DIR
FATA initBuildConfig failed: clibuildconfig cannot be created due to error in creating build context, getAbsBuildContext failed: exit status 128 
```

Now this error sadly did not explain a lot to me and I was confused because the kompose version didn't even change.
After some debugging I figured out that the error comes from the git command executed in this function: https://github.com/kubernetes/kompose/blob/main/pkg/transformer/openshift/utils.go#L34

Now the git version is something that has indeed changed in the container and executing the command directly inside the container gives me the following output:
```
fatal: detected dubious ownership in repository at '{{ PATH }}'
To add an exception for this directory, call:
	git config --global --add safe.directory {{ PATH }}
```

I think it would be useful to return this information as part of the kompose error output.
Similiary if I was to provide a compose file with a build context that does not point to a git repository I would be hit with the exact same error:
`FATA initBuildConfig failed: clibuildconfig cannot be created due to error in creating build context, getAbsBuildContext failed: exit status 128`
instead of receiving the much more useful information from git:
```
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

I can create a PR to adjust the error output. I've seen there are 2 more git commands in the same file that could possibly also benefit from more output. Please let me know if you see any issues with my approach.

Even if no PR ends up being merged here it might help someone that you can google for this error now.
The error did not return any results in google so far.

### Actual Behavior

better error output

### Steps To Reproduce

_No response_

### Kompose Version

```Text
same behaviour in 1.22.0 and 1.34.0
```

### Docker-Compose file

_No response_

### Anything else?

_No response_

cortex3:
#### What type of PR is this?

/kind feature

improves error output of git commands see #1957 for context.
Example:
current output:
```
FATA initBuildConfig failed: clibuildconfig cannot be created due to error in creating build context, getAbsBuildContext failed: exit status 128
```
with my PR:
```
FATA initBuildConfig failed: clibuildconfig cannot be created due to error in creating build context, getAbsBuildContext failed: fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

Fixes #1957

### DIFF
--- a/pkg/transformer/openshift/utils.go
+++ b/pkg/transformer/openshift/utils.go
@@ -1,6 +1,7 @@
 package openshift
 
 import (
+	"github.com/pkg/errors"
 	"os/exec"
 	"strings"
 )
@@ -34,12 +35,16 @@ func GetImageTag(image string) string {
 func GetAbsBuildContext(context string) (string, error) {
 	cmd := exec.Command("git", "rev-parse", "--show-prefix")
 	cmd.Dir = context
-	out, err := cmd.Output()
+	var out strings.Builder
+	var stderr strings.Builder
+	cmd.Stdout = &out
+	cmd.Stderr = &stderr
+	err := cmd.Run()
 	if err != nil {
-		return "", err
+		return "", errors.New(stderr.String())
 	}
 	//convert output of command to string
-	contextDir := strings.Trim(string(out), "\n")
+	contextDir := strings.Trim(out.String(), "\n")
 	return contextDir, nil
 }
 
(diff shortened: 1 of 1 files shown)


cortex3:
Hi @cdrage,
I noticed that there are already unit tests for the openshift utils.go. They are just a bit hidden away because they have been included in the openshift_test.go: https://github.com/kubernetes/kompose/blob/922f5aaeefceb9162d32094768e968511552062e/pkg/transformer/openshift/openshift_test.go#L256

The tests are still passing but let me know if you want me to extend them in any way.

cdrage:
> Hi @cdrage, I noticed that there are already unit tests for the openshift utils.go. They are just a bit hidden away because they have been included in the openshift_test.go:
> 
> https://github.com/kubernetes/kompose/blob/922f5aaeefceb9162d32094768e968511552062e/pkg/transformer/openshift/openshift_test.go#L256
> 
> The tests are still passing but let me know if you want me to extend them in any way.

Honestly, tests are passing and this adds crucial information for the openshift section (which needs improvement to be honest...), this LGTM.

PR #1967, opened 2024-12-09
Add new instruction to permanently enable completion in the [bash|zsh|fish] shell

### LINKED ISSUE #1966: [BUG] Update Kompose completion
### Expected Behavior

kompose completion generate a script that introduces a completion for the shell but this feature does not work for new instance of shell. 

### Actual Behavior

Added a suggestion to add these commands to shell configuration file to enable completion permanently.

### Steps To Reproduce

_No response_

### Kompose Version

```Text
1.34
```

### Docker-Compose file

_No response_

### Anything else?

_No response_

utkarsh-singh1:
#### What type of PR is this?

/kind bug

kompose completion generate a script that introduces a completion for the shell but this feature does not work for new instance of shell.

Fixes #1966

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -18,13 +18,13 @@ var completion = &cobra.Command{
 Auto completion supports bash, zsh and fish. Output is to STDOUT.
 
 # Enable auto-completion in bash  
-source <(kompose completion bash)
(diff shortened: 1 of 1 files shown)


cdrage (changes requested):
Sorry, but all this PR does is add comments?

PR #1975, opened 2024-12-18
Replaced y with ✓

alf:
#### What type of PR is this?

/kind documentation

Replaces usage of `y` in the conversion table with ✓.

`y` is not a valid state according to the glossary.
It is hard to spot the difference between `x` and `y`, so fixing this will make it less likely that reads the wrong thing from the table.

If `y` does not mean ✓ then `y` should be added to the glossary.
Again, since these marks are hard to differentiate I hope this won't be necessary.


cdrage:
Good catch @alf 

Sign the k8s CLA and I can merge this in! :)

TessaIO (approved):

alf:
My pleasure. Small things also matter.

PR #1993, opened 2025-02-14
Fix case of FromAs

gisostallenberg:
#### What type of PR is this?
bug

This fixes a warning in newest docker builder

```
FromAsCasing: 'as' and 'FROM' keywords' casing do not match
```

Fixes #


kundan2707 (commented):
seems good as per docker image file command format

Issue #1999, opened 2025-03-14
Compose `env_file` should result in a `envFrom` in deployment


robertaistleitner:
### What would you like to be added?

I want `env_file` entries in Docker Compose to result in `envFrom` entries in the Deployment instead of single entries using `configMapKeyRef` in `env`.

### Why is this needed?

At the moment a env_file is properly converted to a ConfigMap which is nice and all the contained environment variables are then added to containers `env` using `configMapKeyRef`s. I think this is overcomplicating the deployment file since one could simply use `envFrom` and load the full `ConfigMap` to the container envs.

This should look like this:
```
envFrom:
    - configMapRef:
        name: example-env
```

Is there anything I didn't think of which could prevent this change?

robertaistleitner:
This is fixed with #2001

PR #2001, opened 2025-03-17
Fix ConfigMap name clash if env_file name is equal

### LINKED ISSUE #2000: [BUG] `ConfigMap` name clash if `env_file` name is equal
### Expected Behavior

All the used env_file configs should result in a separate `ConfigMap` file, even though the file name is the same. For this to work, the path of the file needs to be included in the file name to be unique.

### Actual Behavior

If you use two `env_file`s in docker-compose which are located in different folders and have the same name (e.g. `one/custom.env` and `two/custom.env`) this results in only one `ConfigMap` file in the rendered result. One env file is lost in the rendering.

### Steps To Reproduce

_No response_

### Kompose Version

```Text
1.35.0 (9532ceef3)
```

### Docker-Compose file

```YAML
services:
    nginx:
        image: nginx
        env_file:
            - ./one/custom.env
            - ./two/custom.env
```

### Anything else?

_No response_

robertaistleitner:
#### What type of PR is this?

/kind bug

Fixes naming collisions if env files have same filename in different folders.

Fixes #2000

### DIFF
--- a/pkg/transformer/kubernetes/k8sutils.go
+++ b/pkg/transformer/kubernetes/k8sutils.go
@@ -973,12 +973,9 @@ func GetContentFromFile(file string) (string, error) {
 // FormatEnvName format env name
 func FormatEnvName(name string, serviceName string) string {
 	envName := strings.Trim(name, "./")
-	// only take string after the last slash only if the string contains a slash
-	if strings.Contains(envName, "/") {
-		envName = envName[strings.LastIndex(envName, "/")+1:]
-	}
 
-	envName = strings.Replace(envName, ".", "-", -1)
+	// replace all non-alphanumerical characters with dashes to have a unique envName (env filename could be used multiple times)
+	envName = regexp.MustCompile(`[^a-zA-Z0-9]`).ReplaceAllString(envName, "-")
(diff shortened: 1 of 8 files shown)


robertaistleitner:
There is probably one more issue I just found: ordering of the environment variables is also incorrect (in docker compose the order of env_files describe the priority of env vars to be applied, which is also the case for Kubernetes as far as I know).

cdrage:
> There is probably one more issue I just found: ordering of the environment variables is also incorrect (in docker compose the order of env_files describe the priority of env vars to be applied, which is also the case for Kubernetes as far as I know).

Please open an issue :)

You are right, it's not 1-1 with the spec and it should order it correctly in case one of the env var's overrides another.

Either way, thanks for the contribution and once the tests pass, this already LGTM to me.

robertaistleitner:
@cdrage since this fix also corresponds to the issue in #1999 and the ordering issue, I added it here as well.

cdrage (approved):
Sorry for the delay! Just catching up to kompose.

This LGTM, thanks for the awesome contribution! The code is great, there are tests added, and I can confirm on my end it works well with Kubernetes.

/lgtm
/approve

PR #2098, opened 2026-04-13
fix: normalize underscores to dashes in Service metadata.name

### LINKED ISSUE #1897: [BUG] underscores in compose service name are not replaced with dashes in kubernetes Service metadata.name
### Expected Behavior

`kompose convert` should replace underscores with dashes in kubernetes Service metadata.name when compose service name contains underscores

```
apiVersion: v1
kind: Service
metadata:
  labels:
    io.kompose.service: myservice-api
  name: myservice-api  # THIS IS OK
spec:
  ports:
    - name: "8000"
      port: 8000
      targetPort: 8000
  selector:
    io.kompose.service: myservice-api
```

### Actual Behavior

It generates:

```
apiVersion: v1
kind: Service
metadata:
  labels:
    io.kompose.service: myservice-api
  name: myservice_api  # <-- THIS IS NOT VALID name
spec:
  ports:
    - name: "8000"
      port: 8000
      targetPort: 8000
  selector:
    io.kompose.service: myservice-api
```

`metadata.name` is not valid (it contains underscore instead of dash)

### Steps To Reproduce

_No response_

### Kompose Version

```Text
1.34.0
```

### Docker-Compose file

_No response_

### Anything else?

_No response_

vigneshakaviki:
## Summary

Fixes #1897

When a Docker Compose service name contains underscores (e.g., `my_web_app`), kompose correctly normalizes the map key and most resource names to use dashes (`my-web-app`), but the `serviceConfig.Name` field retains underscores because it's set via `parseResourceName()` which uses `normalizeContainerNames()` (lowercase only, no underscore replacement).

This causes the Kubernetes **Service** `metadata.name` to contain underscores (`my_web_app`) while other resources like **Deployments** correctly use dashes (`my-web-app`).

## Root Cause

In `dockerComposeToKomposeMapping()`, `serviceConfig.Name` is initially set from `parseResourceName()` which calls `normalizeContainerNames()`. The map key is set from `normalizeServiceNames()` (which replaces underscores with dashes), but `serviceConfig.Name` is never updated to match.

## Fix

Set `serviceConfig.Name` to the result of `normalizeServiceNames()` so all generated Kubernetes resources use consistent, DNS-valid names.

## Before (stock kompose)

```
$ kompose convert -f docker-compose.yaml --stdout
# Service gets: name: my_web_app    ← invalid, has underscores
# Deployment gets: name: my-web-app ← correct
```

## After (patched)

```
$ kompose convert -f docker-compose.yaml --stdout
# Service gets: name: my-web-app    ← correct
# Deployment gets: name: my-web-app ← correct
```

## Test Plan

- Added `TestServiceConfigNameNormalized` unit test that:
  - Creates a compose project with service name `my_service`
  - Verifies `serviceConfig.Name` is `my-service` after mapping
  - Confirmed test **fails** without the fix, **passes** with it
- All existing tests pass (`go test ./...`)
- End-to-end verification with `kompose convert --stdout`

### DIFF
--- a/pkg/loader/compose/compose.go
+++ b/pkg/loader/compose/compose.go
@@ -587,10 +587,12 @@ func dockerComposeToKomposeMapping(composeObject *types.Project) (kobject.Kompos
 			return kobject.KomposeObject{}, err
 		}
 
-		// Log if the name will been changed
-		if normalizeServiceNames(name) != name {
-			log.Infof("Service name in docker-compose has been changed from %q to %q", name, normalizeServiceNames(name))
+		// Normalize the service name for Kubernetes (e.g. replace underscores with dashes)
+		normalizedName := normalizeServiceNames(name)
+		if normalizedName != name {
+			log.Infof("Service name in docker-compose has been changed from %q to %q", name, normalizedName)
 		}
+		serviceConfig.Name = normalizedName
 
 		serviceConfig.Configs = composeServiceConfig.Configs
 		serviceConfig.ConfigsMetaData = composeObject.Configs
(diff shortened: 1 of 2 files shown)


cdrage:
Hi @vigneshakaviki the main blocker would be the commit message, any way you can modify it to pass? :) thanks

cdrage:
Closing as commit message hasn't been updated. Feel free to open up another PR if possible! Just cleaning up some PR's on this repo. thanks.

Issue #1070, opened 2018-09-14
Duplicate container ports if different host ports point to same container port


dkarlovi:
If you have something like this:

```yaml
version: '3'
services:
    test:
        image: alpine
        ports:
        - '8000:80'
        - '80:80'
```

you end up with this:

```yaml
- apiVersion: extensions/v1beta1
  kind: Deployment
  spec:
    template:
      spec:
        containers:
        - image: alpine
          name: test
          ports:
          - containerPort: 80
          - containerPort: 80
```

which creates problems with usage on K8s.

PR #2103, opened 2026-05-09
fix: dedupe exposed ports with normalized protocols

immanuwell:
Fixes a tiny duplicate-port edge case

Repro before this patch:

```yaml
services:
  app:
    image: nginx
    ports:
      - "80"
    expose:
      - "80"
      - "8080"
```

`ports` comes from compose-go as `tcp`, while `expose` defaults to `TCP`, so the dedupe key missed and emitted `80/TCP` twice. not great.

This normalizes protocols before checking the key, so `80/TCP` is kept once and `8080/TCP` still gets added.

Related history: #1070 / #1223 fixed another duplicate-port path; this is the ports+expose version.

Tested:

```sh
go test ./...
```


HaraldNordgren (approved):
Makes sense!

immanuwell:
@HaraldNordgren just rebaised this PR and now also all CI checks should be green
PTAL

HaraldNordgren (approved):

cdrage (changes requested):
So sorry been a delay in reviews!

Can you please add a unit test as well? 

These are typically added to this folder: https://github.com/kubernetes/kompose/tree/main/script/test/fixtures

cdrage:
I have not seen any updates and am just cleaning up the PR's, if you are able to update it, please re-open! :) many thanks!

PR #2104, opened 2026-05-13
fix: ensure unique mountPath when service has multiple secrets

### LINKED ISSUE #1894: [BUG] Multiple secrets for single service result in non-unique "mountPath" error
### Expected Behavior

Multiple secrets for a single service are correctly translated.

### Actual Behavior

"kubectl" reports:
```
Error from server (Invalid): error when creating "service-deployment.yaml": Deployment.apps "service" is invalid: spec.template.spec.containers[0].volumeMounts[1].mountPath: Invalid value: "/run/secrets": must be unique
```

### Steps To Reproduce

1. Prepare first secret: `echo -n 'password1' > textfile1`
2.Prepare second secret: `echo -n 'password2' > textfile2`
3. `kompose convert -f compose.yaml`
5. `minikube start`
6. `kubectl apply -f .`

### Kompose Version

```Text
1.34 (cbf2835db)
```

### Docker-Compose file

```YAML
name: "test"

secrets:
  "secret1":
    file: "textfile1"
  "secret2":
    file: "textfile2"

services:

  "service":
    image: "hello-world:latest"
    ports:
      - "80:80"
    secrets:
      - "secret1"
      - "secret2"
```

### Anything else?

_No response_

fru1tworld:
#### What type of PR is this?

/kind bug

When a service references multiple secrets, compose-go normalizes
short-syntax `secrets: [a, b]` to long-syntax with
`Target="/run/secrets/<name>"`. `getSecretPathsLegacy` then stripped the
filename and used only the parent directory as `mountPath`, so both
secrets ended up with `mountPath: /run/secrets` and `kubectl apply` failed
with `volumeMounts[*].mountPath: must be unique`.

Use the full target as the mountPath. SubPath still selects the file
within the secret data, so the file lands at the same location.

Fixes #1894

### DIFF
--- a/pkg/transformer/kubernetes/kubernetes.go
+++ b/pkg/transformer/kubernetes/kubernetes.go
@@ -931,38 +931,31 @@ func (k *Kubernetes) getSecretPathsLegacy(secretConfig types.ServiceSecretConfig
 	// docker-compose results, but some people might depend on this behavior so this is kept here for compatibility.
 	// See https://github.com/kubernetes/kompose/issues/1280 for more details.
 
-	var itemPath string // should be the filename
-	var mountPath = ""  // should be the directory
-	// if is used the short-syntax
+	var itemPath, mountPath string
+
+	// short-syntax (no target): place at /run/secrets/<source>
 	if secretConfig.Target == "" {
-		// the secret path (mountPath) should be inside the default directory /run/secrets
 		mountPath = "/run/secrets/" + secretConfig.Source
-		// the itemPath should be the source itself
 		itemPath = secretConfig.Source
-	} else {
-		// if is the long-syntax, i should get the last part of path and consider it the filename
-		pathSplitted := strings.Split(secretConfig.Target, "/")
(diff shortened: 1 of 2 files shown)


cdrage:
> #### What type of PR is this?
> /kind bug
> 
> #### What this PR does / why we need it:
> When a service references multiple secrets, compose-go normalizes short-syntax `secrets: [a, b]` to long-syntax with `Target="/run/secrets/<name>"`. `getSecretPathsLegacy` then stripped the filename and used only the parent directory as `mountPath`, so both secrets ended up with `mountPath: /run/secrets` and `kubectl apply` failed with `volumeMounts[*].mountPath: must be unique`.
> 
> Use the full target as the mountPath. SubPath still selects the file within the secret data, so the file lands at the same location.
> 
> #### Which issue(s) this PR fixes:
> Fixes #1894

@fru1tworld Are you able to rebase please? If so, tests will then pass! Many thanks

cdrage:
Closing for now until rebased  (please re-open when you do)

PR #2105, opened 2026-06-15
fix(ci): pin GitHub Actions to full-length commit SHAs

cdrage:
fix(ci): pin GitHub Actions to full-length commit SHAs

The kubernetes org requires all actions to be pinned to full commit SHAs
rather than version tags.

This PR hopefully passes all the tests now.


cdrage:
Merging this in order to pass tests for other MR's 👍

PR #2108, opened 2026-08-23
Reject a volume that is only a name and an access mode

### LINKED ISSUE #2109: [BUG] kompose convert panics on a volume that is only a name and an access mode
### Expected Behavior

`kompose convert` should report a malformed volume entry as an error and exit, the way it already does for other bad volume strings.

### Actual Behavior

It panics with `index out of range [-1]` and prints a Go stack trace. Exit code is 2.

```
panic: runtime error: index out of range [-1]

goroutine 1 [running]:
github.com/kubernetes/kompose/pkg/transformer.parseVolume({0x398b7b3b6350, 0x8})
	/kompose/pkg/transformer/utils.go:112 +0x4bc
github.com/kubernetes/kompose/pkg/transformer.ParseVolume({0x398b7b3b6350, 0x8})
	/kompose/pkg/transformer/utils.go:72 +0x60
github.com/kubernetes/kompose/pkg/loader/compose.ParseVols({0x398b7b5a01a0?, 0x1, 0x0?}, {0x398b7b3b63b0, 0x3})
	/kompose/pkg/loader/compose/compose.go:928 +0xf8
github.com/kubernetes/kompose/pkg/loader/compose.retrieveVolume(...)
	/kompose/pkg/loader/compose/compose.go:903 +0xc8
github.com/kubernetes/kompose/pkg/loader/compose.handleVolume(...)
	/kompose/pkg/loader/compose/compose.go:837 +0xb8
github.com/kubernetes/kompose/pkg/loader/compose.dockerComposeToKomposeMapping(...)
	/kompose/pkg/loader/compose/compose.go:609 +0xb94
github.com/kubernetes/kompose/pkg/app.Convert(...)
	/kompose/pkg/app/app.go:221 +0xbc
```

### Steps To Reproduce

1. Save the compose file below.
2. Run `kompose convert --stdout`.
3. It panics instead of printing an error.

`myvol:rw`, `myvol:z` and `myvol:Z` do the same thing. Well-formed entries such as `myvol:/data:ro` and `/host:/data:ro` are unaffected.

### Kompose Version

```Text
1.38.0 (HEAD)
```

built from `main` at c2dd9614fd5026b8f6cb95d61bba802626088b13.

### Docker-Compose file

```YAML
services:
  web:
    image: nginx
    volumes:
      - "myvol:ro"

volumes:
  myvol:
```

### Anything else?

`parseVolume` in `pkg/transformer/utils.go` takes the volume name off the front of the split list and a trailing `rw`/`ro`/`z`/`Z` off the back, then reads the container path with `volumeStrings[len(volumeStrings)-1]`. For `myvol:ro` those two steps consume the whole list, so the index goes to -1.

The entry is not usable in the first place. Compose reads the second field as the container target rather than as an access mode, and the engine then rejects it:

```
Error response from daemon: invalid volume specification: 'repro-vol_myvol:ro:rw': invalid mount config for type "volume": invalid mount path: 'ro' mount path must be absolute
```

So an error is the right outcome here, and `parseVolume` already returns `invalid volume format` for a bare `ro` via the same length check one branch earlier. #2108 adds the equivalent check after the access mode is stripped.

arpitjain099:
#### What type of PR is this?

/kind bug

`parseVolume` takes the volume name off the front of the split list and a trailing `rw`/`ro`/`z`/`Z` off the back, then reads the container path with `volumeStrings[len(volumeStrings)-1]`. When the entry is just a name and an access mode, both ends consume the whole list and the index goes to -1, so `kompose convert` panics instead of reporting a bad volume:

```yaml
volumes:
  - "myvol:ro"
```

```
panic: runtime error: index out of range [-1]
```

This adds a length check after the access mode is stripped, and returns the `invalid volume format` error that `parseVolume` already returns for a bare `ro` through the same check one branch earlier.

Fixes #2109

Ran `kompose convert --stdout` on the compose file above, against `main` and against this branch. `main` panics and exits 2 with the trace in #2109. This branch exits 0, and `myvol:/data:ro` still converts to the same `mountPath: /data` and `readOnly: true` as before.

Direct calls into `parseVolume` on `main`, covering the whole shape:

```
vol:ro           -> PANIC: runtime error: index out of range [-1]
vol:rw           -> PANIC: runtime error: index out of range [-1]
vol:z            -> PANIC: runtime error: index out of range [-1]
vol:/data:ro     -> name="vol" host="" container="/data" mode="ro" err=<nil>
/host:/data:ro   -> name="" host="/host" container="/data" mode="ro" err=<nil>
ro               -> name="ro" host="" container="" mode="" err=invalid volume format: ro
```

Two tests. One covers all four access-mode spellings and asserts an error rather than a panic, and it fails on `main` with the panic above. The other pins four well-formed volumes to their exact parsed fields, so the new check cannot start rejecting valid input.

`go build ./...`, `go vet ./pkg/transformer/` and `go test ./pkg/transformer/...` are clean, including the `kubernetes` and `openshift` sub-packages.

One thing I ran into while checking the end to end behavior, which I have deliberately left alone. `handleVolume` in `pkg/loader/compose/compose.go` calls `errors.Wrap` on the error from `retrieveVolume` and discards the result, so a volume parse error never reaches the user and the service converts with the volume silently dropped. That is why the run on this branch exits 0 rather than printing the error. It predates this change, and fixing it properly means giving `handleVolume` an error return and threading it through `dockerComposeToKomposeMapping`, so it felt wrong to fold into a panic fix. Glad to open a separate issue and PR for it if you would like.


cdrage:
> `parseVolume` takes the volume name off the front of the split list and a trailing `rw`/`ro`/`z`/`Z` off the back, then reads the container path with `volumeStrings[len(volumeStrings)-1]`. When the entry is just a name and an access mode, both ends are consumed and nothing is left:
> 
> ```yaml
> volumes:
>   - "myvol:ro"
> ```
> 
> ```
> panic: runtime error: index out of range [-1]
> ```
> 
> Ran it against `main` before and after. Before:
> 
> ```
> vol:ro           -> PANIC: runtime error: index out of range [-1]
> vol:rw           -> PANIC: runtime error: index out of range [-1]
> vol:z            -> PANIC: runtime error: index out of range [-1]
> vol:/data:ro     -> name="vol" host="" container="/data" mode="ro" err=<nil>
> /host:/data:ro   -> name="" host="/host" container="/data" mode="ro" err=<nil>
> ro               -> name="ro" host="" container="" mode="" err=invalid volume format: ro
> ```
> 
> so the valid forms were fine and only the name-plus-mode shape blew up. Note the last line: a bare `ro` already returns the invalid-format error, via the same length check one branch earlier. This adds the equivalent check after the access mode is stripped, and returns that same error rather than a new one.
> 
> Two tests. One covers all four access-mode spellings and asserts an error rather than a panic. The other pins four well-formed volumes to their exact parsed fields, so the new check cannot start rejecting valid input. The first fails on `main` with the panic above.
> 
> `go build ./...`, `go vet ./pkg/transformer/` and `go test ./pkg/transformer/...` are clean, including the `kubernetes` and `openshift` sub-packages.

Can you please update with what issue this closes (and if no issue, please open), and follow the https://github.com/kubernetes/kompose/blob/main/.github/PULL_REQUEST_TEMPLATE.md PR template for this PR? Many thanks!

arpitjain099:
Done, thanks. There was no existing issue for it, so I opened #2109 with the compose file, the stack trace and the version, and rewrote the description to follow the template with `Fixes #2109`.

While putting the issue together I checked what Compose itself does with `myvol:ro`, which I had not before. It reads the second field as the container target rather than as an access mode, and the engine then refuses it with `mount path must be absolute`, so an error really is the right outcome here rather than kompose trying to make sense of it.

One thing I flagged in the special notes rather than changing: `handleVolume` calls `errors.Wrap` on the error from `retrieveVolume` and drops the result, so with this fix applied the bad volume is silently omitted from the output and `kompose convert` still exits 0. Fixing that means giving `handleVolume` an error return and threading it up through `dockerComposeToKomposeMapping`, which seemed like more than a panic fix should carry. Happy to open it separately if you want it addressed.

/kind bug

PR #2119, opened 2026-08-26
fix: normalize Service and StatefulSet.Spec.ServiceName to match metadata.name

### LINKED ISSUE #1897: [BUG] underscores in compose service name are not replaced with dashes in kubernetes Service metadata.name

sooraj-sky:
#### What type of PR is this?

/kind bug

`InitSvc` in `pkg/transformer/kubernetes/kubernetes.go` set `Service.ObjectMeta.Name` from `service.Name`, which only ever gets lowercased (`parseResourceName` -> `normalizeContainerNames`), instead of the `name` parameter the function already receives — which is the correctly normalized map key (`normalizeServiceNames`: lowercase + underscores/dots converted to dashes). Every other resource constructor (`Deployment`, `DaemonSet`, `CronJob`, and even `StatefulSet`'s own `ObjectMeta.Name`) already uses that `name` parameter, so `Service` was the one inconsistent case: a compose service named `my_web_app` correctly produced a Deployment named `my-web-app`, but the Service stayed `my_web_app` — not a valid Kubernetes resource name (RFC1123).

While tracing this I found the identical bug in `InitSS`: `StatefulSet.Spec.ServiceName` was also set from `service.Name`. This matters because `Spec.ServiceName` must match the headless Service's `metadata.name` for the StatefulSet's governing-service reference to actually resolve. Verified concretely: before this fix, a StatefulSet-controlled service named `my_db_service` produced a headless Service named `my_db_service` and `Spec.ServiceName: my_db_service` (consistent with each other, but both invalid), while the StatefulSet's own name was already correctly `my-db-service` — a real mismatch waiting to happen once the Service name got fixed on its own. Both are now `name`.

Fixes #1897

There's a prior PR (#2098) for this same issue that was approved in review but closed for a stale commit message / missing CLA signature (the maintainer invited a clean resubmission). I did not reuse that PR's approach — it mutated `serviceConfig.Name` globally in the compose loader, and its own CI was failing (Build/Test/lint all red), which suggests that field is read elsewhere in ways that weren't accounted for. This PR is narrower: only the two spots that build Kubernetes name fields (`InitSvc`, `InitSS`) now use the `name` parameter they already had; nothing else changes.

Verified locally end-to-end, not just via unit tests: built the binary, ran `kompose convert` against both a plain-Service compose file and a `--controller=statefulset` compose file with underscore service names, confirmed the pre-fix output (checked out the parent commit) versus post-fix output for both cases. Then ran the full CI-equivalent gate: `./script/check-gofmt.sh`, `go vet ./pkg/...`, `go test -short -race -cover ./...` across every package, and `make test-cmd` (the integration script against the built binary) — diffed against a baseline run on unmodified `main` to confirm this change introduces zero new failures (a handful of `test-cmd` cases fail identically on both, due to a missing local `dyff` binary and docker-image-build-dependent fixtures, unrelated to this change).

Added `TestServiceNameNormalized` and `TestStatefulSetServiceNameNormalized` to `kubernetes_test.go` covering both fixed paths.

AI-assisted contribution: I used an AI coding assistant (Claude) to help trace the root cause, apply the fix, and write the regression tests. I reviewed, tested, and verified all of the above myself before submitting.


cdrage:
> #### What type of PR is this?
> /kind bug
> 
> #### What this PR does / why we need it:
> `InitSvc` in `pkg/transformer/kubernetes/kubernetes.go` set `Service.ObjectMeta.Name` from `service.Name`, which only ever gets lowercased (`parseResourceName` -> `normalizeContainerNames`), instead of the `name` parameter the function already receives — which is the correctly normalized map key (`normalizeServiceNames`: lowercase + underscores/dots converted to dashes). Every other resource constructor (`Deployment`, `DaemonSet`, `CronJob`, and even `StatefulSet`'s own `ObjectMeta.Name`) already uses that `name` parameter, so `Service` was the one inconsistent case: a compose service named `my_web_app` correctly produced a Deployment named `my-web-app`, but the Service stayed `my_web_app` — not a valid Kubernetes resource name (RFC1123).
> 
> While tracing this I found the identical bug in `InitSS`: `StatefulSet.Spec.ServiceName` was also set from `service.Name`. This matters because `Spec.ServiceName` must match the headless Service's `metadata.name` for the StatefulSet's governing-service reference to actually resolve. Verified concretely: before this fix, a StatefulSet-controlled service named `my_db_service` produced a headless Service named `my_db_service` and `Spec.ServiceName: my_db_service` (consistent with each other, but both invalid), while the StatefulSet's own name was already correctly `my-db-service` — a real mismatch waiting to happen once the Service name got fixed on its own. Both are now `name`.
> 
> #### Which issue(s) this PR fixes:
> Fixes #1897
> 
> #### Special notes for your reviewer:
> There's a prior PR (#2098) for this same issue that was approved in review but closed for a stale commit message / missing CLA signature (the maintainer invited a clean resubmission). I did not reuse that PR's approach — it mutated `serviceConfig.Name` globally in the compose loader, and its own CI was failing (Build/Test/lint all red), which suggests that field is read elsewhere in ways that weren't accounted for. This PR is narrower: only the two spots that build Kubernetes name fields (`InitSvc`, `InitSS`) now use the `name` parameter they already had; nothing else changes.
> 
> Verified locally end-to-end, not just via unit tests: built the binary, ran `kompose convert` against both a plain-Service compose file and a `--controller=statefulset` compose file with underscore service names, confirmed the pre-fix output (checked out the parent commit) versus post-fix output for both cases. Then ran the full CI-equivalent gate: `./script/check-gofmt.sh`, `go vet ./pkg/...`, `go test -short -race -cover ./...` across every package, and `make test-cmd` (the integration script against the built binary) — diffed against a baseline run on unmodified `main` to confirm this change introduces zero new failures (a handful of `test-cmd` cases fail identically on both, due to a missing local `dyff` binary and docker-image-build-dependent fixtures, unrelated to this change).
> 
> Added `TestServiceNameNormalized` and `TestStatefulSetServiceNameNormalized` to `kubernetes_test.go` covering both fixed paths.
> 
> AI-assisted contribution: I used an AI coding assistant (Claude) to help trace the root cause, apply the fix, and write the regression tests. I reviewed, tested, and verified all of the above myself before submitting.

Looks like it's failing a test at the moment, if you can fix and rerun.

Please also add claude as an author since this was AI assisted

PR #2124, opened 2026-09-16
fix: do not emit duplicate PVC artifacts for StatefulSets

### LINKED ISSUE #2090: StatefulSet with kompose.volume.type=persistentVolumeClaim generates duplicate PVC artifacts
## Describe the bug

When a service uses `kompose.controller.type: statefulset` together with `kompose.volume.type: persistentVolumeClaim`, `kompose convert` generates three artifacts for the same volume — two of which are redundant and cause a broken deployment:

1. `volumeClaimTemplates` inside the StatefulSet ✅ (correct — this is what the pod uses)
2. A standalone `<volume>-persistentvolumeclaim.yaml` ❌ (orphaned — no consumer, stays `Pending` forever)
3. A `spec.template.spec.volumes[]` entry referencing the same volume ❌ (redundant — conflicts with `volumeClaimTemplates`)

The pod binds to the PVC auto-created by `volumeClaimTemplates` (e.g. `pgdata-postgresql-0`), while the standalone `pgdata` PVC sits `Pending` indefinitely with no consumer.

## To Reproduce

Docker Compose service with these labels:

```yaml
services:
  postgresql:
    image: postgres:13-alpine
    volumes:
      - pgdata:/var/lib/postgresql/data
    labels:
      kompose.controller.type: statefulset
      kompose.volume.size: 1Gi
      kompose.volume.type: persistentVolumeClaim

volumes:
  pgdata:
```

Run:

```sh
kompose convert -c -f docker-compose.yaml -o my-chart
```

## Expected behavior

For a StatefulSet, the volume should be managed exclusively by `volumeClaimTemplates`. kompose should:
- Emit `volumeClaimTemplates` in the StatefulSet ✅
- **Not** emit a standalone `pgdata-persistentvolumeclaim.yaml`
- **Not** add a `spec.template.spec.volumes[]` entry for the same volume (it is owned by `volumeClaimTemplates`)

## Actual behavior

Three files/entries are generated for the same volume. Deploying to Kubernetes results in an orphaned PVC stuck in `Pending` and the StatefulSet failing due to the conflicting `spec.volumes` entry.

## Workaround

```sh
# Remove redundant spec.volumes entry
yq -i 'del(.spec.template.spec.volumes[] | select(.name == "pgdata"))' \
    my-chart/templates/postgresql-statefulset.yaml

# Delete orphaned standalone PVC manifest
rm -f my-chart/templates/pgdata-persistentvolumeclaim.yaml
```

## Environment

- kompose version: `1.38.0 (a8f5d1cbd)`
- Kubernetes: kind cluster
- OS: Linux

## Root cause (suspected)

`ConfigVolumes` in `pkg/transformer/kubernetes/kubernetes.go` unconditionally emits standalone `PersistentVolumeClaim` objects regardless of the target controller type. The StatefulSet path correctly generates `volumeClaimTemplates`, but `ConfigVolumes` is not suppressed for the StatefulSet case, leading to both artifacts being emitted simultaneously.

This was introduced with StatefulSet support in [PR #1452](https://github.com/kubernetes/kompose/pull/1452).

carterpewpew:
#### What type of PR is this?

/kind bug

A service converted to a StatefulSet owns its persistent volume through `volumeClaimTemplates`, but conversion also emitted two redundant artifacts for the same volume: a standalone PersistentVolumeClaim, which has no consumer and
stays `Pending`, and a `spec.template.spec.volumes[]` entry bound to the same claim, which conflicts with the claim `volumeClaimTemplates` creates.

This skips both for StatefulSets, and keeps the volumes that are not claim-backed in the pod template, so a container never mounts a volume the object does not declare.

The check reads the objects that were actually generated rather than the requested controller type. A service with `restart: "no"` or `on-failure` and no `--controller` flag is converted to a bare Pod or a CronJob even when the `kompose.controller.type` label asks for a StatefulSet, and neither of those has `volumeClaimTemplates` to own the claim, so both keep the standalone claim and the pod volume that references it.

Fixes #2090

This PR was made with AI Assistance, but was tested and verified manually.

Three tests are added, each checked to fail without the production change it covers:

- `TestStatefulSetDoesNotDuplicatePVC` fails on the current code with the
  symptoms from the report.
- `TestPodFromRestartPolicyKeepsStandalonePVC` fails if the controller type is
  read from the label rather than from the generated objects.
- `TestStatefulSetKeepsNonClaimVolumes` fails if non-claim volumes are dropped
  from the pod template.


cdrage:
Thank you!! This helps a lot. LGTM :) Let's get it merged in.

